PR #20, opened 2016-09-26
moved Network call off UI thread.

### LINKED ISSUE #17: Network Call blocks UI
Network call for reply is blocking UI thread, this needs to be fixed.

gogeta95:
UI is not getting blocked anymore on sending new message.
fixes #17.


shivenmian:
@gogeta95 great, using Asynctask there is a better optimisation. merging.

Issue #15, opened 2016-09-26
Soft keyboard get open automatically


amitiwary999:
After opening app soft keyboard get open automatically

nehagup:
i would like to take up this issue.

amitiwary999:
I already started working on it

vikiCoder:
Sorry, but I did solv and sent pull request...

amitiwary999:
I also solved that issue but not able to send PR.It always show denied permission.

daminisatya:
Please assign the issue. who ever wants to solve it. and then send the pull request. Inorder to avoid conflict of work. This would waste a lot of energy.

daminisatya:
I dont see any assignees to this issue. so let's follow the best practice.

daminisatya:
@vikiCoder Please resolve the comments which i have gave to your pull request. Thank you.

PR #29, opened 2016-09-27
Fixes #15 :Solve soft keyboard get open automatically

amitiwary999:
Fixes issue #15 :Added some code in onPause method in mainactivity and also made some changes in activity_main.xml so that it solve soft keyboard get open automatically issue.

PR #33, opened 2016-09-27
Fixes issue #15: solved soft keyboard get open automatically 

amitiwary999:
Fixes issue #15 to solve this issue added some code in mainactivity to hide soft keyboard so that it will open only when user touch edittext.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/MainActivity.java
@@ -14,6 +14,7 @@ import android.text.TextUtils;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.WindowManager;
 import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
@@ -119,7 +120,10 @@ public class MainActivity extends AppCompatActivity {
     @Override
     protected void onResume() {
         super.onResume();
-        etMessage.requestFocus();
+        //etMessage.requestFocus();
+        getWindow().setSoftInputMode(
+                WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_HIDDEN
+        );
     }
 
     @OnClick({R.id.iv_image, R.id.btn_send})


daminisatya:
+1

daminisatya:
#29

app/src/main/java/org/fossasia/susi/ai/MainActivity.java
```diff
@@ -119,7 +120,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
     @Override
     protected void onResume() {
         super.onResume();
-        etMessage.requestFocus();
+        //etMessage.requestFocus();
```
gogeta95: why was this line commented out?

gogeta95: @daminisatya please review the PR before merging as this one breaks the keyboard popping up feature instead of solving it.

gogeta95: I see what happened, I wanted the keyboard to open while opening the app, while the OP didn't. so the question is do we need keyboard to open automatically on startup or not?

amitiwary999: In chat app user never want that soft keyboard open automatically.

the-dagger: Technically the user won't want the keyboard to popup as he enters the chat app. (As it is in the case of Whatsapp or Allo)

gogeta95: We can keep it as it is then, this is just fine.

daminisatya: @gogeta95 Yes sir will do so.

gogeta95: Just call me Saurabh :)

sudheesh001: @gogeta95 I think this patch is perfectly fine and the merge is justified, no chat app really brings up the keyboard automatically and that shouldn't be the way it should work, commenting it out and leaving a comment there in case we change the way the interaction should take place should be the right step. 

This is the correct fix to the breaking user experience caused in the patch to open up automatically.

gogeta95:
@sudheesh001, you're absolutely correct, it was just a misunderstanding of
requirements, no worries! :)

PR #35, opened 2016-09-27
fixes #31: multiple performance issues with splash screen.

### LINKED ISSUE #31: Multiple Issues with splash screen.
- [x] Splash activity extends Activity, should extend Appcompativity.
- [x] Image is loaded in XML, should be loaded with an image library to resize according to screen size.
- [x] layout can be improved by removing RelativeLayout.

@the-dagger what library should we use for image loading? Picasso or glide? i prefer glide because it handles image resizing effectively.

gogeta95:
Adressed #31 .
* Activity now extends AppcompatActivity
* Image is now loaded dynamically.
* Improved layout by decreasing one level.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,5 +38,8 @@ dependencies {
     */
     compile 'com.jakewharton:butterknife:8.4.0'
     apt 'com.jakewharton:butterknife-compiler:8.4.0'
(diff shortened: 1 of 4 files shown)


daminisatya:
+1 looks good.

PR #45, opened 2016-09-29
Make link for url in message

djKooks:
How about give a link for url in message? Like when I input "check google.com", the link has been set to google.com. Please look on.


daminisatya:
+1

PR #47, opened 2016-09-29
Fix #46: Compile view lib binary

### LINKED ISSUE #46: Compile only a binary for the external Library
I think we should only compile the binary for the View Library until we rewrite, which is available [here](https://github.com/himanshu-soni/ChatMessageView#installation).
Project management and gradle compilation is faster after proguard and easier maintenance.

addiittya2006:
The Code is perfectly adaptable and Clean after this. Any View Changes can be pushed to the library itself.


addiittya2006:
@the-dagger Review and Merge.

the-dagger:
LGTM

Issue #49, opened 2016-09-29
If user write multiple line message then edittext cover the large portion of UI


amitiwary999:
If user line message with multiple line then edittext box will cover the large portion of UI.

sagarchoudhary96:
yaa that's why i want to make it single line.... :)

the-dagger:
Similar issue is already opened in #48

sagarchoudhary96:
i have fixed this one too .. just sending the pull request

PR #52, opened 2016-09-29
Fixes issue #49 Added inputtype textmultiline with max line 4.

amitiwary999:
Fixes issue #49 If user write a messahe of multiple line edittext box will show onlt last 4 line so that editbox will not cover the large portion of UI.


daminisatya:
@amitiwary999 Please provide us with a screenshot.

daminisatya:
@the-dagger Is the number 4, being decided?

the-dagger:
@daminisatya I think that it is more of a UX choice.
A normal question shouldn't exceed 4 lines and hence it is used imo.

Issue #48, opened 2016-09-29
EditTextBox on Enter key text sending


sagarchoudhary96:
when we press enter key, it goes to second line. we can make it that on pressing enter msg get sent

daminisatya:
+1

the-dagger:
@sagarchoudhary96 Sounds good, do you want to work on this?

sagarchoudhary96:
yeah sure .. i would love to work on it :)

sagarchoudhary96:
@the-dagger  okay i will see to it too

amitiwary999:
But My work is different.User may write multiline text without pressing enter key and it will hide large portion of UI.

the-dagger:
@amitiwary999 Yeah saw it just now.
Will reopen your issue

gogeta95:
How about we make it like WhatsApp?
Limit of  4 lines max and "enter is send" is user preference?

gogeta95:
This solves both #48 and #49.

addiittya2006:
#52 A preference toggle must be added in settings to use 'Enter as Send' as most queries to Susi are short.

amitiwary999:
@sagarchoudhary96  if you are working on it then try to solve issue #49  also . Otherwise I will try to solve both issue.

sagarchoudhary96:
@amitiwary999 i'll see to it

amitiwary999:
@sagarchoudhary96 are you working on it. @the-dagger I would  like to work on this. At last you decide whose code you want to pull.

sagarchoudhary96:
@amitiwary999 I have completed it .. i will update pull request as soon i reach home :)

sagarchoudhary96:
For issue 49 as written that User may write multiline text without pressing enter key and it will hide large portion of UI. If enter key as send preference is enabled then the text will be a single line only and if it is disable than it will go to multiline and since it is chatbot what i think is no user will be ever asking questions more tan 2-3 lines at max.  if question is more than 7-8 lines then only it will cover a large portion of ui which i don't think will be a case ...

amitiwary999:
If you know  human being then you know that they can ask anything.

sagarchoudhary96:
@daminisatya can we know close this issue :)

PR #60, opened 2016-09-30
Fixes issue #49 and #48 Enter key work as send button and at a time user can see last three line

amitiwary999:
Fixes issue #49 and #48 Now enter key work as send button and in case of multiline message at a time user can see only last three lines so it will not hide large portion of UI.


gogeta95:
Hey, @sagarchoudhary96 has already sent a PR for #48. Please push code only for #49 which is assigned to you. Thanks.

Issue #68, opened 2016-09-30
Use Retrofit + GSON for networking and JSON Parsing.


gogeta95:
we should use retrofit for network calls and GSON for JSON parsing (instead of manual) to make the code more cleaner and understandable.

gogeta95:
@the-dagger @daminisatya  will move ahead with #63 after this is done, as current parsing only gives us a text message, we need more parameters in message object.

PR #74, opened 2016-10-01
addressed #68: added retrofit + GSON.

gogeta95:
added retrofit and GSON for networking and JSON parsing respectively.


gogeta95:
@sudheesh001 updated the PR, please review.

PR #82, opened 2016-10-02
Issue 48 Fixed. Added Enter Key as Send Feature.

sagarchoudhary96:
User can select between to choose whether to use Enter key to send
message or function as normal .


daminisatya:
@sagarchoudhary96 looks good

sagarchoudhary96:
@daminisatya thanks :)

PR #100, opened 2016-10-07
Fixes #73 : speech to text updated 

### LINKED ISSUE #73: Using speech to text to take the input
how it would be to use speech to text input in the app..
i would like to work on that....

mayank408:
fixes #73 , now the user can give input to the app by his or her voice  ...


mariobehling:
Please provide a description with info "fixes 'xyz'". Thank you.

mariobehling:
Ok, thank you. Let's try this.

PR #114, opened 2016-10-09
Fixes #113 UI improvement

### LINKED ISSUE #113: UI improvement
Like in whatsapp , when the user has not typed any message there is a single button for speech to text , as soon as user types some message the button gets changed to send button . I would like to implement this.

mayank408:
@mariobehling Now the single button can work as send as well as speech to text input.
whatsapp image 2016-10-09 at 5 16 59 pm
whatsapp image 2016-10-09 at 5 16 59 pm 1


jig08:
@mayank408 Looks great.
Squash your commits into one following the naming pattern before we can merge the changes.

PR #117, opened 2016-10-09
fixes issue #108 improving search UI

amitiwary999:
fixes issue #108 Now if search text found then only text background colour will change.


mariobehling:
Thank you.

Issue #130, opened 2016-10-10
Search selection needs improvement


nikimullick:
- [ ] Once the **search** is done and closed, the text **remains selected**(coloured).It doesn't go away!! Unless a new search is started.

- [ ] Selection varies on User and Susi's side-The search selection is coloured(black background) on the user side but the text comes in black on susi's side.It must be the same kind on both sides
Similar issue was there #108 but seems like it was solved only on the user's end of chat.

![screenshot_2016-10-10-17-20-44-588_org fossasia susi ai 1](https://cloud.githubusercontent.com/assets/20987692/19235224/06c6dcb2-8f0e-11e6-94f7-4076234d7a96.png)

![screenshot_2016-10-10-17-28-36-848_org fossasia susi ai 1](https://cloud.githubusercontent.com/assets/20987692/19235513/c2d0ff4a-8f0f-11e6-9fd0-7dbefee9e4b6.png)

jig08:
@nikimullick Nice find. Fix this pls.

jig08:
Also see if PR https://github.com/fossasia/susi_android/pull/129 doesn't clash with your fix.

nikimullick:
okay .I'll see to that

amitiwary999:
@nikimullick have you fixed it? #108 was assigned to me and I forgot to change susi message background.I solved it so I am sending PR.

nikimullick:
cool @amitiwary999  you can solve **2**. I'll do the other one in search.

jig08:
@nikimullick @amitiwary999 Edited the initial issue to accommodate both.

amitiwary999:
@nikimullick I am not getting your first point.

nikimullick:
@jig08 seems like today isn't my day :( That search issue is solved in the new commit .

the-dagger:
@nikimullick Point 1 isn't solved yet.
@amitiwary999 's PR solves only Point 2 of the issue.
You can work on this if you want to .

amitiwary999:
@the-dagger point 1 is not a issue it is working fine in my android phone.

nikimullick:
@the-dagger point 1 is already resolved.I don't know by whom but it is not there now in the latest commit.However, a little while ago it was an issue .Now,its working fine

amitiwary999:
@nikimullick always try to update your code with master branch of fossasia/susi_android.

nikimullick:
I did but meanwhile some other commits happened

[...]

PR #133, opened 2016-10-10
fixes issue #108 and #130 improving search UI

amitiwary999:
fixes issue #108 and #130 in my previous commit I changed only user-message and forgot to change susi-message background colour so solved that issue.Now if search text found then only text background colour will change.


the-dagger:
@amitiwary999 Could you please attach an after/before screenshot?
Also please do it everytime you make UI changes.
Cheers.

amitiwary999:
@the-dagger you can see #130 for before part and here is after change screenshot.from now onward i will attach screenshot if required.
screenshot_2016-10-10-19-00-39

the-dagger:
Thanks!

PR #142, opened 2016-10-11
Fixes #141: Managed Hard Coded Strings in strings.xml

### LINKED ISSUE #141: Manage Hard Coded Strings in strings.xml
Everywhere, contributors are using hard coded strings because there is no rule to manage strings.ml. 
This issue for managing the strings resources  in strings.xml.

therajanmaurya:
fixes #141 Managed Hard Coded Strings in strings.xml.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -7,7 +7,6 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
-import android.graphics.Color;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import android.net.ConnectivityManager;
@@ -16,8 +15,8 @@ import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.speech.RecognizerIntent;
 import android.support.design.widget.CoordinatorLayout;
-import android.support.design.widget.FloatingActionButton;
 import android.support.design.widget.Snackbar;
+import android.support.v4.content.ContextCompat;
 import android.support.v4.view.MenuItemCompat;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
@@ -44,6 +43,7 @@ import android.widget.Toast;
 
 import org.fossasia.susi.ai.R;
 import org.fossasia.susi.ai.adapters.recyclerAdapters.ChatFeedRecyclerAdapter;
+import org.fossasia.susi.ai.helper.Constant;
 import org.fossasia.susi.ai.helper.DateTimeHelper;
 import org.fossasia.susi.ai.model.ChatMessage;
 import org.fossasia.susi.ai.rest.ClientBuilder;
@@ -55,7 +55,6 @@ import java.util.Locale;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
-import butterknife.OnClick;
 import io.realm.Case;
 import io.realm.Realm;
 import io.realm.RealmResults;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -200,13 +196,12 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                         //SharePreference to store image
                         shre = PreferenceManager.getDefaultSharedPreferences(this);
                         SharedPreferences.Editor edit = shre.edit();
-                        edit.putString("image_data", encodedImage);
+                        edit.putString(Constant.IMAGE_DATA, encodedImage);
```
the-dagger: Ideally, only those string resources should go into Strings.xml which are visible to the user.
Here, "image_data" is used as a key and not visible to the user, so there is no need of moving it to Strings.xml

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -218,9 +213,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                         cropCapturedImage(selectedImageUri);
                     } catch (ActivityNotFoundException aNFE) {
                         //display an error message if user device doesn't support
-                        String errorMessage = "Sorry - your device doesn't support the crop action!";
-                        Toast toast = Toast.makeText(this, errorMessage, Toast.LENGTH_SHORT);
-                        toast.show();
+                        showToast(getString(R.string.error_crop_not_supported));
```
the-dagger: These are the type of messages we need to move inside Strings.xml

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -300,33 +300,33 @@ public void cropCapturedImage(Uri picUri){
         cropIntent.putExtra("return-data", true);
         startActivityForResult(cropIntent, CROP_PICTURE);
     }
-    public void setChatBackground(){
+
+    public void setChatBackground() {
         shre = PreferenceManager.getDefaultSharedPreferences(this);
-        String previouslyChatImage = shre.getString("image_data", "");
+        String previouslyChatImage = shre.getString(Constant.IMAGE_DATA, "");
```
the-dagger: Same thing here again, only move those strings into Strings.xml which are visible to the user.

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
therajanmaurya: @the-dagger, So that I have made Constant.java a helper to manage Key values.
It's not in string.xml

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
therajanmaurya: @the-dagger, So that I have made Constant.java a helper to manage Keys values.
It's not in strings.xml

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
therajanmaurya: It's already in the strings.xml

the-dagger: Yeah, I know.
I was just specifying that these are the type of message we need and not the earlier one.

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
the-dagger: Ok, I missed the constants.java.
Looks good now. :+1:

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
therajanmaurya: you mean others strings that i have put in strings.xml that should nor be there.

the-dagger: Nope, its all good. :+1:

jig08:
@the-dagger See to it that the commits are named following the format `Fixes #<issue number> <commit message>` mentioned in the README itself before merging the changes, we all need to keep up with dev and maintenance flow of the project.

the-dagger:
Missed it, will make sure of it from now onwards.

PR #165, opened 2016-10-12
Fixes #155: Back-End Configuration and fail-over 

### LINKED ISSUE #155: Back-End Configuration and fail-over
The susi api is currently hard-coded in https://github.com/fossasia/susi_android/blob/master/app/src/main/java/org/fossasia/susi/ai/rest/ClientBuilder.java
We should have a system variable for this.

Furthermore, the access to the susi api should have a fail-over process. That means we should have the ability to declare a set of back-end addresses, not only one.

To do:
- declare a variable susi_service_count (default: 2)
- declare two variables:
 * susi_service_0 = api.asksusi.com
 * susi_service_1 = kaskelix.com:4000
- when the back-end service 0 is accessed but ends in a time-out, try service 1 and so on

therajanmaurya:
### Fixes issue #
Made base url abstract, So we can change base url, when ever we want.

### Changes : 
Added BaseUrl.java for managing Base Url. Like we will add a method to change base url in SharedPreferences and refresh the Retrofit Object by calling ClientBuilder.createSusiService()
and it's done, do not need to do anything else.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -49,6 +49,7 @@ dependencies {
     //Retrofit
     compile 'com.squareup.retrofit2:retrofit:2.1.0'
     compile 'com.squareup.retrofit2:converter-gson:2.1.0'
+    compile "com.squareup.okhttp3:logging-interceptor:3.3.1"
 
     //preference UI
     compile 'com.takisoft.fix:preference-v7:24.2.1.0'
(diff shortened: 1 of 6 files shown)


therajanmaurya:
@the-dagger Please look, So I will make changes if required.

the-dagger:
@therajanmaurya Looks Good.
Also good job by adding comments in the code. :slightly_smiling_face:

therajanmaurya:
Thanks.

gogeta95:
@the-dagger multiple issues with this PR.
1. breaks the singleton pattern in Retrofit API.
2. why add the logging interceptor, it has nothing to do with the issue being solved.
3. this contains only functionality of changing the baseURLs anytime. does not provide fail safe.
 In case one service goes down, we need to immideately access new service(baseURL) as @Orbiter wrote: 
>declare two variables:
* susi_service_0 = api.asksusi.com
* susi_service_1 = kaskelix.com:4000
when the back-end service 0 is accessed but **ends in a time-out, try service 1 and so on**

PR #179, opened 2016-10-13
Fixes #39, Merged with Login test and added 2 new test cases for MainActivity

### LINKED ISSUE #39: Addition of stricter unit tests.
Current unit tests on Travis just check for successful build but does not check whether various app functionality is working correctly or not.

We may add support for checking basic app functionalities like sending messages using androidTest framework so that errors like in #34  can be avoided.

abhi007tyagi:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ script:
   - echo no | android create avd --force -n test -t $ANDROID_TARGET --abi $ANDROID_ABI
   - emulator -avd test -no-audio -no-window &
   - android-wait-for-emulator
+#  - adb shell input keyevent 26 &
   - adb shell input keyevent 82 &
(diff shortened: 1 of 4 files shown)


abhi007tyagi:
image

Output at my end !!

abhi007tyagi:
Test result on Real Device

image

abhi007tyagi:
One thing I can think of is I am using LoginActivity as Rule. If MainActivity is used, then MainActivity is created first and then all the test cases are run. Since on creation MainActivity requires Login, one cannot run any test case. 
So, in order to run test cases in Main Activity, I have first run Login Test (under @Before) and then test each test case for MainActivity. These should and are running as after login Main Activity is opened automatically.

PR #184, opened 2016-10-13
Fixes #95 and Fixes #107 added pie chart,feature to send messages afterconnecting to internet

### LINKED ISSUE #95: Add support for Pie charts in Susi Response
Susi server is well ahead giving response in various forms, one such visualization is pie charts. The app must provide support to process the pie data and shoot out a pie chart as a result. Susi server tells you what kind of data its going to give. 
Susi webclient uses high charts to display the pie charts

### LINKED ISSUE #107: no response when there is no internet
if a user sends a message when there is no internet then he should get relevant answers once he is connected to internet

iit2014086:
Fixes issue #95 #107

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     compile "com.android.support:design:$support_lib_version"
 
     compile 'me.himanshusoni.chatmessageview:chat-message-view:1.0.3'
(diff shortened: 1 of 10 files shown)


daminisatya:
@iit2014086 is the server issue resolved?

iit2014086:
@daminisatya no it hasn't but i checked at every step with hardcoded values ;) ...i was sure it will work and added that part too.

PR #187, opened 2016-10-14
fixes #156: Added the logic to change base url whenever SocketTimeoutException will come

### LINKED ISSUE #156: remote configuration for susi_service
Every time the susi app starts up, it reads an internal configuration for the Susi back-end server location as defined in https://github.com/fossasia/susi_android/issues/155

This configuration should be overwritten with the values as given in
http://config.asksusi.com/config_susi.json

This should happen only one time when the susi App is started.

The purpose of this remote configuration is the ability to change the back-end scaling without the need to deploy new susi versions with new back-end configuration.

To do:
- read the JSON from http://config.asksusi.com/config_susi.json
- overwrite the susi_service array as created in https://github.com/fossasia/susi_android/issues/155 with the "susi_service" value from http://config.asksusi.com/config_susi.json

therajanmaurya:
### Fixes issue #
Added Logic to change base Url whenever SocketTimeoutException will come
### Changes : 
Added method in BaseUrl to change base url.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,6 +21,11 @@
         <activity
             android:name=".activities.SplashScreen"
             android:theme="@style/AppTheme.NoActionBar">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
         </activity>
         <activity
             android:name=".activities.SettingsActivity"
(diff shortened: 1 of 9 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,11 @@
         <activity
             android:name=".activities.SplashScreen"
             android:theme="@style/AppTheme.NoActionBar">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
```
gogeta95: why did you enabled the splash screen again?

therajanmaurya: because we need to the fetch Susi Base URLs list every time whenever app will open, we cannot wait to any SocketTimeoutException come and this is the best place to fetch.
we can not wait for SocketTimeoutException because if any SocketTimeoutException came during fetching susi response we cannot make new request to susi server to ask which server is running, this current is not.

gogeta95: you need to write that in changes section.

therajanmaurya: Yeah will do that.

PR #202, opened 2016-10-15
Fixes #167 repeated code

### LINKED ISSUE #167: repeated code
in chatFeedRecyclerAdapter the same OnLongClick method is implemented repeatedly...there will be many different types of holders and need to be implemented many times...so instead make the Recycleradapter implement OnLongClickListener and for every chatMessageview set onlongclicklistener as class itself instead of creating new listener.

piyush6348:
Fixes issue #167 

Changes: made ChatViewHolder and MapViewHolder classes to implement onLongClick,modified accordingly in ChatFeedRecyclerAdapter so as to avoid onLongClick declaration for all holders

Screenshots for the change: [Add Screenshots or changes.]

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recyclerAdapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recyclerAdapters/ChatFeedRecyclerAdapter.java
@@ -126,7 +126,7 @@ public class ChatFeedRecyclerAdapter extends RecyclerView.Adapter<RecyclerView.V
             case USER_MESSAGE:
                 view = inflater.inflate(R.layout.item_user_message, viewGroup, false);
                 return new ChatViewHolder(view, USER_MESSAGE);
-           case SUSI_MESSAGE:
+            case SUSI_MESSAGE:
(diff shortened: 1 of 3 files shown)


mayank408:
@piyush6348 you have just cut the code from recycleview class and paste in the their respective holders . How does it helped it reduction of repeated code.?? The no of lines are still the same.

iit2014086:
@piyush6348  dont make the view holders implement onlongclick instead make the adapter implement it.

mayank408:
@piyush6348 You are using copy method twice in each holder , where previously we are using it only once . More lines of repeated code.
Also there are two more holders for pie chart and link , we cant keep making seperate code for all.

piyush6348:
@iit2014086 @mayank408 I was facing some issues
when I made recyclerview class to implement onlongclick

mayank408:
@piyush6348 Long Clicks were already implemented in Recycleview class. But you took them to respective holder. What you can do is make a method inside the recycleview that can handle longclicks for each holder.
thankyou

the-dagger:
Closing this, as this is not exactly a meaningful improvement to the number of lines in the codebase.

PR #204, opened 2016-10-15
fixes #182 SignUp screen after sign up is sucessfull , snackbar added

### LINKED ISSUE #182: SignUp screen after sign up is sucessfull
### Expected Behaviour

After the successful sign up the activity should go back on the log in page
### Actual Behaviour

It only shows a successful signup toast and user has to press back to get to login page 
### Would you like to work on the issue?

Yes I would like to solve this issue. 
I will work on it.

nimit95:
Fixes issue #182 

Changes: now after successful sign up screen is directed to login screen. Snackbar shows instead of Toast message in case of error.


nimit95:
I am not able to find why travis build failed .

jig08:
@the-dagger Could have a look here ?

the-dagger:
Please sync your fork and send the Pull Request again so that it can pass CircleCI tests.

PR #218, opened 2016-10-16
Fixes #206 : Audio Focus issues resolved.

### LINKED ISSUE #206: Requesting audio focus when there is sound in background.
During text to speech , susi app should request audio focus . So there is no overlapping of sound.

mayank408:
Fixes issue #206 
Changes:
Audio Focus added to text to speech.


mayank408:
@jig08 @daminisatya Pls merge this as I have two more pr to make for different issues.

PR #223, opened 2016-10-16
Fix #222 highlighting search result in map and link message

### LINKED ISSUE #222: Highlighting search result in chatmessage doesn't work if query is part of message containing link or map
**Actual Behaviour**

Highlighting search result in chatmessage doesn't work if query is part of message containing link or map. 

**Expected Behaviour**

query searched should be highlighted in link message.

**Steps to reproduce it**

send a link from your side. "thank you <link>" you get in reply for now.
search for  thank/you or any substring of link , its not highlighted.

Yes

viveksb007:
Fixes issue #222 

Changes: search query is highlighted in messages with map and link.

Screenshots for the change:


daminisatya:
+1

PR #224, opened 2016-10-16
Fixes #196 : Back button issue resolved

### LINKED ISSUE #196: Issue on pressing back button in setting.
### Expected Behaviour
The app on pressing back button should go to main activity.
### Actual Behaviour
App is crashing on pressing back button in settings.

### Would you like to work on the issue?
Yes

mayank408:
Now on pressing back button on settings it goes to main activity.


the-dagger:
I don't think that this is the right way to fix this bug. 
We should just call finish on the setting activity rather than creating an intent of mainActivity., as it can cause more bugs later on.

PR #254, opened 2016-10-18
Fixes #237 : Chat box UI improved.

### LINKED ISSUE #237: Improving chat box UI
**Actual Behaviour**

If there are even only six messages it covers the whole screen of chat box.
**Expected Behaviour**

As in popular message app the chat box can be made a little concise to provide more space in a single view.

- [x] Place the timestamp in par with the text. 

- [x] Make the chat bubbles flatter. Reducing the spacing between each bubble to a reasonable amount.

Yes

mayank408:
Fixes issue #237 
Screenshots for the change: 
image


the-dagger:
Don't hardcode the settings, please move them to ```dimens.xml```

daminisatya:
@mayank408 No where close to the whatsapp UI. I wont get that excitement when I see clumsy chat bubbles. We need think of it more. cause this is the major visible part of the app.

PR #280, opened 2016-10-22
Fixed #277: Deactivate speech outputs for map URLs

### LINKED ISSUE #277: Deactivate voice output for map URL locations
**Actual Behaviour**

When speech output is enabled, Susi speaks out the long map URL on questions like "Where is xyz". 

**Expected Behaviour**

When Susi says "Here is a map", Susi should not mention the URL of the map.

NiharG15:
Fixes issue #277 

Changes:  
 - Rename voicereply() to voiceReply() and change signature to include isMap.
 - If the reply is a map, remove the URL from the spoken reply.

Screenshots for the change: 
NA


mariobehling:
Good job.

PR #282, opened 2016-10-22
 Fixes #276 : Speech Output issue resolved.

### LINKED ISSUE #276: Enable Voice Output if user speaks to app
**Actual Behaviour**

Currently the user needs to enable voice output of Susi in the settings. Even if the user uses voice to communicate.

**Expected Behaviour**

* [x]  The output of voice should be enabled if users uses voice input.
* [x]  There should be an additional option in settings deactivate/activate Voice Output if user uses "Speech to Text", e.g. "At speech input enable speech out"
* [ ] The currently existing option "Speech Output" should be changed to "Speech Output Always On"

mayank408:
Fixes issue #276

Changes:Now the speech output is given only when the user gives input via speech to text.
It can be disabled in the settings as well.

Screenshots for the change: 
whatsapp image 2016-10-22 at 5 10 02 pm

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -111,6 +111,7 @@ public class MainActivity extends AppCompatActivity {
     RealmResults<ChatMessage> chatMessageDatabaseList;
     Boolean micCheck;
     private SearchView searchView;
+    public  Boolean check;
     private Menu menu;
     private int pointer;
     private RealmResults<ChatMessage> results;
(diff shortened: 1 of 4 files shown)


daminisatya:
@mayank408 What is the third option? not clear

mayank408:
@daminisatya At speech input enable speech output.

daminisatya:
We need to think of a simpler chat setting options. To a new person, i feel it's kinda of difficult.

mayank408:
@daminisatya I will change it in my next pr for issue #284 . Please merge this for now. Thanks

mariobehling:
Please read the issue description carefully. One point is missing.

mayank408:
@mariobehling I did not get the meaning of 3 point . Could you explain what exactly I have to do.

daminisatya:
@mayank408 If a reviewer's job is to just merge it, Then we would have suggested you to push code directly to master. The problem is not with the code merging and updating the existing issue in the next PR. We want the best to go in. Thank you

mayank408:
@mariobehling Is it alright now?

whatsapp image 2016-10-22 at 8 11 19 pm

mayank408:
@daminisatya Ok I will ensure the best possible practice.

mariobehling:
Yes, good. Please set following as standard:

* Enter as send: Enabled
* Use mic for input: Enabled
* At speech input enable speech output: Enabled
* Speech output always on: Disabled

Of course use can change the setting.

mayank408:
@mariobehling Standard values updated

mayank408:
@mariobehling @daminisatya Is i have to make any changes to make it merged.

mariobehling:
Thank you. Let's try this out a bit. Good progress.

Issue #267, opened 2016-10-20
Simplify design of login/signup screens


mariobehling:
Currently there are different colors on the login screens and the app does not yet have a color profile for all screens.

The app UI should be kept as simple as possible for now. Login screens similar to the Facebook messenger can be an example.

Therefore please implement as follows:
- [ ] Make all background screens for login white similar to FB messenger
  ![screenshot from 2016-10-20 01 56 35](https://cloud.githubusercontent.com/assets/1583873/19541801/7a85b800-9669-11e6-8691-7bbc6a535a0b.png)
- [ ] Order the text and buttons similar to the FB sign up page
  ![ahfpl92](https://cloud.githubusercontent.com/assets/1583873/19541825/a6771198-9669-11e6-82f6-78eaf1f53405.png)

amitiwary999:
@mariobehling I would like to work on this.

daminisatya:
Assigned @amitiwary999

amitiwary999:
@mariobehling please let me know if I need any changes.
![-10-20-19-11-31](https://cloud.githubusercontent.com/assets/14003244/19562321/79edc73a-96f9-11e6-8f67-7783d2f808dc.png)
![screenshot_2016-10-20-19-12-55](https://cloud.githubusercontent.com/assets/14003244/19562331/813c6dde-96f9-11e6-87e7-1e143cb45ba9.png)

mariobehling:
Thank you. Please follow the order and way the buttons are displayed of Facebook and colors similar to Messenger e.g.
* [ ] Change color of font to gray
* [ ] no sign up button, but - text (like in FB messenger)
* [ ] no different colors for forgot password etc.
* [ ] Square buttons
* [ ] Also follow original description

amitiwary999:
@mariobehling review?
![screenshot_2016-10-20-21-53-24](https://cloud.githubusercontent.com/assets/14003244/19568638/fa197060-970f-11e6-87b9-7642c052593f.png)
![screenshot_2016-10-20-21-53-32](https://cloud.githubusercontent.com/assets/14003244/19568639/fa1b3828-970f-11e6-8f08-5662fb8feba4.png)

mariobehling:
@amitiwary999 We value your work, but let's not get too much into detailed discussions. 

Please read carefully the description and comments in the issue, e.g. background should be white.

mariobehling:
There are also button that can be ticked. These buttons are there to understand if all details have been implemented. Buttons like that:
[...]

PR #286, opened 2016-10-22
fixes issue #267 simplify design of login/signup screen

amitiwary999:
Fixes issue #267 simplify design of login/signup screen


mariobehling:
Thank you!

Issue #289, opened 2016-10-23
App crashing on startup


mariobehling:
**Actual Behaviour**

I was able to use the app a short time and login first. I tested the voice feature and the first crash happened. After that, I am not able to start Susi. It is crashing on startup.

**Expected Behaviour**

No crashes.

amitiwary999:
@mariobehling can you post the error log because it is not crashing on my device.

mariobehling:
@amitiwary999 Did you try with sound?

I am out at the moment. Can we have some more automated reports/logs of crashes? I opened an issue for this https://github.com/fossasia/susi_android/issues/290

NiharG15:
@mariobehling I'm getting a crash due to a null log message. Sending a PR to fix it.

amitiwary999:
@mariobehling Yeah I got it. 
FATAL EXCEPTION: Thread-19745
                                                                      Process: org.fossasia.susi.ai, PID: 18222
                                                                      java.util.NoSuchElementException
                                                                          at java.util.LinkedList.getFirstImpl(LinkedList.java:538)
                                                                          at java.util.LinkedList.getFirst(LinkedList.java:530)
                                                                          at org.fossasia.susi.ai.activities.MainActivity.computeOtherMessage(MainActivity.java:535)
                                                                          at org.fossasia.susi.ai.activities.MainActivity.access$1100(MainActivity.java:85)
                                                                          at org.fossasia.susi.ai.activities.MainActivity$computeThread.run(MainActivity.java:875)

PR #291, opened 2016-10-23
Fix crash if throwable message is null

NiharG15:
Possibly related to issue #289 

Changes: 
 - The app crashes if t.getLocalizedMessage() returns null. Hence, add a null check.

Screenshots for the change: 
NA


mariobehling:
Thank you.

PR #295, opened 2016-10-23
Fixes #284 added alert dialog for logout screen

### LINKED ISSUE #284: Issue with logout
**Actual Behaviour**

When we click on logout is does not ask for confirmation.
**Expected Behaviour**

When we click on logout it should ask for confirmation.

Yes ofcourse

madhavrathi:
Fixes issue #284

Changes: Added Alert Dialog Box for Logout Screen

Screenshots for the change: 
screen


mariobehling:
Thank you.

Issue #296, opened 2016-10-23
Add a button to scroll to the end like whatsapp


vikiCoder:
We can add a button in chatwindow to scroll to end like whatsapp. Like shown in screenshot...

![screenshot_2016-10-24-01-38-56](https://cloud.githubusercontent.com/assets/16588032/19629124/49219d18-998b-11e6-8db7-daabb3af03a7.jpg)

I would like to work on this. In fact I have started working on....

PR #298, opened 2016-10-23
Fixes issue #296

vikiCoder:
Fixes issue #296 : Added a button to scroll to the end like whatsapp

Added a floating button to scroll to the end of messages. This button will only be visible if the last visible message is older then the third-last message.


jig08:
@vikiCoder Squash your commits into one with the name `Fixes #296 : <commit message>`, follow the best dev practices. Cheers !

PR #301, opened 2016-10-25
 Fixes #299 : Chat dots issue resolved.

### LINKED ISSUE #299: Issue with chat dots.
**Actual Behaviour**

When the user is not connected to internet and send a message , in reply the susi message keeps on showing the chat dots until the net gets connected.

**Expected Behaviour**
If internet in not connected chat bubbles should not show up.

mayank408:
Fixes issue #299

Changes: Now the chat dots dont show up when net is not connected

whatsapp image 2016-10-25 at 10 14 03 am


daminisatya:
Approved!

PR #305, opened 2016-10-25
Fixes #304 Alignment of send button fixed

### LINKED ISSUE #304: Alignment of send button at max height textbox
**Actual Behaviour**

send button shifts up at max height text box

take2

**Expected Behaviour**

It should be unchanged

take1

yes

mejariamol:
Fixes issue #304
Changes: Alignment of send button corrected
Screenshots for the change: 

take3


daminisatya:
Good observation.

PR #313, opened 2016-10-28
Fixes issue #267 simplified ui design

madhavrathi:
Fixes issue #267

Changes: Changed background color of Login Screen

Screenshots for the change: 
screenshot_20161028-142655


mariobehling:
Background color should be white according to issue https://github.com/fossasia/susi_android/issues/267

Unfortunately this PR cannot get merged therefore.

Issue #311, opened 2016-10-27
Error in chat: Invalid access token


mariobehling:
Follow up to https://github.com/fossasia/susi_android/issues/306#issuecomment-256540167

The chat has some errors as below:

@amitiwary999 @mariobehling 

if (!isNetworkConnected()) {
                                        recyclerAdapter.hideDots();
                                        nonDeliveredMessages.addFirst(new Pair(query, id));
                                        Snackbar snackbar = Snackbar.make(coordinatorLayout,
                                                getString(R.string.no_internet_connection), Snackbar.LENGTH_LONG);
                                        snackbar.show();
                                    } else {
                                        realm.executeTransactionAsync(new Realm.Transaction() {
                                            @Override
                                            public void execute(Realm bgRealm) {
                                                long prId = id;
                                                try {
                                                    ChatMessage chatMessage = bgRealm.where(ChatMessage.class).equalTo("id", prId).findFirst();
                                                    chatMessage.setIsDelivered(true);
                                                } catch (Exception e) {
                                                    e.printStackTrace();
                                                }
                                            }
                                        });
                                        **rvChatFeed.getRecycledViewPool().clear();
                                        recyclerAdapter.notifyItemChanged((int) id);
                                        addNewMessage(getString(R.string.error_occurred_try_again), false, false, false, null);**
                                    }
                                }

@Orbiter stated that this problem is caused by an invalid access token. Probably the token must be renewed in some way. Maybe the token is void when the susi_server restarts.

@mayank408 @amitiwary999 Question: Does the android client has still the log-in information and is able to retrieve a new access token automatically in that case? That would be the most appropriate solution for the problem.

mayank408:
@gogeta95 what are your views..

gogeta95:
Saving password in local storage defeats the purpose of token based authentication. The server shouldn't invalidate tokens on each restart, tokens do have an expiry date, we can set the expiry date to 1-2 months and ask user to re login if token has expired.

mariobehling:
@gogeta95 This is not the intended solution. We do not want the user to have to login again. It is not in line with the experience in other apps.

gogeta95:
In that case, we could implement oauth2.0 where we need to manage 2 tokens, access and refresh. However, I have never done that, so we need someone with more knowledge about this process to guide. More info is at: https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/

Orbiter:
[...]

PR #321, opened 2016-11-02
Fixes #309 Show relevant error messages

### LINKED ISSUE #309: Show relevant error messages in chat box
This is a follow up to issue: https://github.com/fossasia/susi_android/issues/306

Currently there is a general error message if something went wrong in Susi "an error ocurred. please try again."

As a matter of fact the server seems to work perfectly fine and it is unclear what the error is. We need a simple way to understand the cause without having to browse through logs at this stage of development.

Please show real error message/reason for errors instead of "there was an error".

@mejariamol @amitiwary999 Please help if possible.

mejariamol:
Fixes issue #309 

Changes: Two errors were identified and shown accordingly:

1. Invalid Access Token

2. Internet Connectivity Problem

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -634,7 +634,7 @@ public class MainActivity extends AppCompatActivity {
                                         });
                                         rvChatFeed.getRecycledViewPool().clear();
                                         recyclerAdapter.notifyItemChanged((int) id);
-                                        addNewMessage(getString(R.string.error_occurred_try_again), false, false, false, null);
+                                        addNewMessage(getString(R.string.error_invalid_token), false, false, false, null);
                                     }
                                 }
                                 if (isNetworkConnected())
@@ -671,7 +671,7 @@ public class MainActivity extends AppCompatActivity {
                                     });
                                     rvChatFeed.getRecycledViewPool().clear();
                                     recyclerAdapter.notifyItemChanged((int) id);
-                                    addNewMessage(getString(R.string.error_occurred_try_again), false, false, false, null);
+                                    addNewMessage(getString(R.string.error_internet_connectivity), false, false, false, null);
                                 }
                                 BaseUrl.updateBaseUrl(t);
                                 computeOtherMessage();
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -634,7 +634,7 @@ public void execute(Realm bgRealm) {
                                         });
                                         rvChatFeed.getRecycledViewPool().clear();
                                         recyclerAdapter.notifyItemChanged((int) id);
-                                        addNewMessage(getString(R.string.error_occurred_try_again), false, false, false, null);
+                                        addNewMessage("Invalid Access Token !", false, false, false, null);
```
jig08: @mejariamol Kindly use `strings.xml` for this purpose.

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -671,7 +671,7 @@ public void execute(Realm bgRealm) {
                                     });
                                     rvChatFeed.getRecycledViewPool().clear();
                                     recyclerAdapter.notifyItemChanged((int) id);
-                                    addNewMessage(getString(R.string.error_occurred_try_again), false, false, false, null);
+                                    addNewMessage("Internet Connectivity Problem !", false, false, false, null);
```
jig08: Here too.

mejariamol:
@jig08 used `string.xml`.

jig08:
@mejariamol Squash the commits too.

mayank408:
@jig08 @mejariamol  I dont think we should give a message like Invalid acess token .. Its our concern not the one that is shown to the user . It will be a bad user experience.

mejariamol:
@mayank408 It is not going to be a permanent thing. At this stage, it would be better if errors shown are more specific until they are fixed.

mejariamol:
@jig08 squashed.

daminisatya:
@mejariamol @jig08 Is there a possibility to make these messages more user friendly? 

```
Invalid Access Token
Internet Connectivity Problem
```

@mejariamol Can you explain about the ```Invalid access token``` message?

mejariamol:
@daminisatya When the expired access token is used in the request, server sends null body in the SusiResponse. At this stage there is no any implementation for renewing expired access token. Issue #311

daminisatya:
👍 cool merging it.

PR #337, opened 2016-11-05
Fixes #336 TTS now says the desired message.

### LINKED ISSUE #336: Speech output says "I don't know" when Browser search is launched.
**Actual Behaviour**

Whenever we ask anything Susi ain't aware of. Susi launches a web search Intent with a message which reads "Conducting a Search.....Launching Browser." but the problem is the speech output is not the same. Speech output says "I don't know?" 

**Expected Behaviour**

Speech output should be "Conducting a Search.....Launching Browser." instead  of "I don't know."

Yes, I would love to. :)

octacode:
Fixes issue #336. TTS now says "Conducting a Search.....Launching Browser." whenever it conducts a web Search.

Changes: Earlier the TTS used to speak "I don't know" whenever it had to launch a web search. But now the TTS says the desired message.

Thanks. :)


octacode:
Thanks for the merge.
@jig08

Issue #340, opened 2016-11-06
Multi line query not supported


octacode:
**Actual Behaviour**
Susi works great when we query Susi in a single line. But as the query turns into a multi-line query. No matter susi knows the answer or not Susi returns a message saying "An error occurred. please try again."

**Expected Behaviour**

If Susi knows the answer it should return an answer and if Susi doesn't knows the answer it should launch a browser search. That is Susi should work the way it works for a single line query.

**Screenshots of the issue**

![error_input](https://cloud.githubusercontent.com/assets/20819968/20037584/8f072aba-a448-11e6-8864-46b1638331ad.png)

**Would you like to work on the issue?**
Yes, I would love to. :)

daminisatya:
Please open a PR, code looks good! 👍

octacode:
Opened the PR #341
Thanks
@daminisatya

PR #341, opened 2016-11-06
Issue #340 MultiLine query support added.

octacode:
Fixes issue #340

Changes: I have used the split function with whose help I have reconstructed the string. Now Susi for Multi Line query is working in the same way as it works for Single Line Query.


octacode:
Yay! Thanks for the merge.
:)

PR #367, opened 2016-11-28
Fixes #364 : Empty messages issue resolved.

### LINKED ISSUE #364: Empty Messages
**Actual Behaviour**

User can send empty messages to susi.

**Expected Behaviour**

No empty messages should be there.

**Steps to reproduce it**

Step 1.) Go to settings,

screenshot_20161128-044137

Step 2.) Disable "use mic as input" setting under mic settings.

screenshot_20161128-044150

Step3.) Now tap space bar and send empty messages.

screenshot_20161128-044226

yes

mayank408:
Fixes issue #364

Changes: Now if the user doesn't type any message and just print space the message will not be sent as send button will get deactivated if the message is null.


mayank408:
@daminisatya commits squashed.

daminisatya:
cool, thank you! Approved.

Issue #354, opened 2016-11-22
Susi crashes on Text to Speech Input and Speech Output


mariobehling:
**Actual Behaviour**

On a Samsung GT-I1919 and Android 4.4.2 the app crashes when I use text to speech and do not get an immediate answer. When restarting the app, it immediately crashes.

**Expected Behaviour**

No crashes.

SuperHaker:
Happened on my Samsung GT-I8552 too. Cleared app data and restarted the app, and it was fixed.
Something to do with resource consumption on phone?

vivekkiran:
I would like to work on this issue! @mariobehling

SuperHaker:
@mariobehling This issue can be closed. :)

mayank408:
@the-dagger Its merged can be closed now :)

PR #374, opened 2016-11-30
Added TTS to another thread

SuperHaker:
Fixes issue #354 

Changes: Added TTS to another thread to reduce load on main thread, which caused the app to crash.

Screenshots for the change:


mariobehling:
Thanks!

Issue #371, opened 2016-11-29
add timezoneOffset to susi API request


Orbiter:
Currently the ```http://api.asksusi.com/susi/chat.json``` api is called without a timezone information. 

**Expected Behaviour**

For time-based services, we need the local time zone of the android device. The time zone must be computed as minutes delta to the GMT norm as minutes. I.e. if your local position is five hours east of GMT, then you must compute a ```timezoneOffset=-300```.
That parameter must be appended to the susi api call. Currently a call is like:
```
http://api.asksusi.com/susi/chat.json?q=what+time+is+it
```
This must become a call like:
```
http://api.asksusi.com/susi/chat.json?timezoneOffset=-60&q=what+time+is+it
```

If your local position is west of GMT, the timezoneOffset must be a positive number. If you are in the east, the timezoneOffset must be a negative number.

**Steps to reproduce it**

Use the android device with your code an ask "what time is it". Then Susi will tell you the exact time for your local time zone

**Would you like to work on the issue?**

Let us know if this issue should be assigned to you or tell us who you think could help to solve this issue.

SuperHaker:
I would like to work on this issue

PR #376, opened 2016-11-30
Added timezoneOffset in susi api call.

SuperHaker:
Fixes issue #371 

Changes: Added a timezoneOffset parameter in SusiService. Calculated the offset before calling the query in MainActivity.
Verified value of timezoneOffset before sending to server.

Screenshots for the change: 

More info:
Calculation for time should be corrected at server side. Right now it exceeds 24 hours and rounds off timeOffset of 30 mins to 1 hour.


jig08:
@SuperHaker 

* Squash all your commits into one and rename it accordingly as "Fixes #<-issue-number-> : <-commit-message->"
* Fix CI issues as well.

thanks !

PR #379, opened 2016-12-01
Fixes #252: Add option to display multiple answer lines

### LINKED ISSUE #252: Add option to display multiple answer lines
**Actual Behaviour**
Currently the app returns one aswer line for each input

**Expected Behaviour**
It should be possible that the app returns several answer lines, each within it's own box. The answer lines may also appear with a given delay

To test this, you need a response from susi_server which produces several lines. This is prepared for the test question consisting of only one word which is "multi3".
Or test with this URL http://api.asksusi.com/susi/chat.json?q=multi3

The `actions` object should then look like:

```
"actions": [
      {
        "expression": "Line 1",
        "type": "answer"
      },
      {
        "delay": 400,
        "expression": "Line 2",
        "type": "answer"
      },
      {
        "delay": 1000,
        "expression": "Line 3",
        "type": "answer"
      }
    ]
```

You can see three answers, two of them have a delay assigned. For each of those answers with an delay, there should be a pause of the delay value in milliseconds before the answer is written.

samsruti:
Fixes #252

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -114,7 +114,14 @@ public class MainActivity extends AppCompatActivity {
     TextView dates;
 
 	private FloatingActionButton fab_scrollToEnd;
-	
+//	 Global Variables used for the setMessage Method
+    private String answer;
+    private boolean ismap, isPieChart = false;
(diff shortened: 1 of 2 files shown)


daminisatya:
@samsruti Screenshot please

daminisatya:
Code looks good!

samsruti:
@daminisatya Here is the screenshot after the implementation.
before_logout

PR #381, opened 2016-12-03
Fixes #371. Added timezoneOffset in susi api call

### LINKED ISSUE #371: add timezoneOffset to susi API request

SuperHaker:
Fixes issue #371 

Changes: Added a timezoneOffset parameter in SusiService. Calculated the offset before calling the query in MainActivity.
Verified value of timezoneOffset before sending to server.

Screenshots for the change:

More info:
Calculation for time should be corrected at server side. Right now it exceeds 24 hours and rounds off timeOffset of 30 mins to 1 hour.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -69,10 +69,12 @@ import java.io.ByteArrayOutputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
+import java.util.TimeZone;
 import java.util.regex.Matcher;
 
 import butterknife.BindView;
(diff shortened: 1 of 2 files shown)


jig08:
@SuperHaker Nice. Thanks, merging this :)

PR #399, opened 2016-12-08
Fixes #360: Issue regarding websearch 

### LINKED ISSUE #360: Show a google search result list for new answer action type "websearch"
**Actual Behaviour**

currently there is a hard-coded rule inside the app which triggers a web search.

**Expected Behaviour**

The app shall not decide on itself to perform a web search. It must only react on the actions as given in the actions object. To perform a web search, the actions object may now contain an action with the type "websearch". In case that such an action appears, use the value "query" inside the action object to get the actual query term for the web search.

The new actions array will probably contain two actions, one with an normal answer type and another with the new websearch type. Both actions must be performed in the same order as given in the actions array

**Steps to reproduce it**

ask susi the following string: "search for linux distibutions", or open http://api.asksusi.com/susi/chat.json?timezoneOffset=-60&q=search+for+linux+distributions. You will get an result json with the following actions array:
```
"actions": [
      {
        "type": "answer",
        "expression": "Here is a web search result:"
      },
      {
        "type": "websearch",
        "query": "linux distributions"
      }
    ]
```
Use the query object to query the google search result, in this case "linux distributions".
The app shall then show two things:
- the line "Here is a web search result:" - as taken from the expression object
- the google search result in the same window

mayank408:
Fixes issue #360

Changes: Added websearch results to the app using Duckduckgo API.

Screenshots for the change: 

whatsapp image 2016-12-08 at 8 09 49 pm

whatsapp image 2016-12-08 at 8 09 49 pm 1

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -860,7 +860,6 @@ public class ChatFeedRecyclerAdapter extends SelectableAdapter implements Messag
     public class getWebSearch extends AsyncTask<Void, Void, Void> {
 
 
-
         @Override
         protected void onPreExecute() {
             super.onPreExecute();
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -3,8 +3,11 @@ package org.fossasia.susi.ai.adapters.recycleradapters;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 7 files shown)


mayank408:
@jig08 Done..

mayank408:
@mariobehling @the-dagger Is their any changes required....

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -746,4 +855,58 @@ public void shareMessage(String message) {
             currContext.startActivity(sendIntent);
         }
     }
-}
\ No newline at end of file
+
+
+    public class getWebSearch extends AsyncTask<Void, Void, Void> {
```
the-dagger: Why are you making network calls using asynctask?
We are already using retrofit, so use it.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/SearchResultsAdapter.java
```diff
@@ -65,8 +63,8 @@ public void onPos(final SourceContent sourceContent, boolean b) {
                         holder.previewImageView.setVisibility(View.VISIBLE);
                         holder.descriptionTextView.setVisibility(View.VISIBLE);
                         holder.titleTextView.setVisibility(View.VISIBLE);
-                        holder.titleTextView.setText(sourceContent.getTitle());
-                        holder.descriptionTextView.setText(sourceContent.getDescription());
+                        holder.titleTextView.setText("cfcfgxfg");
```
the-dagger: Not sure what you are aiming to do here.

app/src/main/java/org/fossasia/susi/ai/helper/HttpHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.ProtocolException;
+import java.net.URL;
+
+public class HttpHandler {
```
the-dagger: Again, use retrofit.

PR #404, opened 2016-12-09
Fixes #400: Issue regarding showing of map in chat.

### LINKED ISSUE #400: Add action type 'map' to display a map within the chat
**Actual Behaviour**

Currently there is no special handling of geolocation information from actions in susi response

**Expected Behaviour**

A new action type "map" has been introduced. If the susi server responds with a geolocation action result, the android app shall show a map for the given coordinates.
Here is an example:
```
"actions": [
      {
        "type": "map",
        "latitude": "52.52436820069531",
        "longitude": "13.41053001275776",
        "zoom": "13"
      }
    ]
```
If such an action occurs, please show a map in the response of the chat. The zoom value represents a zoom level according to openstreetmap zoom levels.

**Steps to reproduce it**

You can get such an action type if you ask susi "where is berlin" (or any other location). This would be the api call:
http://api.asksusi.com/susi/chat.json?timezoneOffset=-60&q=where+is+berlin

Let us know if this issue should be assigned to you or tell us who you think could help to solve this issue.

mayank408:
Fixes issue #400 
Changes:  Action map is added for map response.

Screenshots for the change: 

whatsapp image 2016-12-09 at 2 20 05 pm

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -613,10 +613,13 @@ public class MainActivity extends AppCompatActivity {
                                                     .get(counter).getExpression();
                                             delay = susiResponse.getAnswers().get(0).getActions()
                                                     .get(counter).getDelay();
-                                            String place = susiResponse.getAnswers().get(0).getData()
-                                                    .get(0).getPlace();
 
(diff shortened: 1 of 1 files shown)


mayank408:
@jig08 Is there any changes required..?

mariobehling:
Looks good to me. Thank you.

PR #418, opened 2016-12-12
Fixes #360 : Issue regarding websearch

### LINKED ISSUE #360: Show a google search result list for new answer action type "websearch"

mayank408:
Fixes issue #360

Screenshots for the change: 

whatsapp image 2016-12-12 at 4 07 16 pm

whatsapp image 2016-12-08 at 8 09 49 pm

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'realm-android'
 
 android {
     compileSdkVersion 24
-    buildToolsVersion "24.0.2"
+    buildToolsVersion '25.0.2'
 
     defaultConfig {
         applicationId "org.fossasia.susi.ai"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'realm-android'
 
 android {
     compileSdkVersion 24
-    buildToolsVersion '25.0.2'
+    buildToolsVersion '24.0.2'
 
     defaultConfig {
         applicationId "org.fossasia.susi.ai"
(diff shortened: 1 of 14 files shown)


mayank408:
@the-dagger Have a look pls..

mayank408:
@the-dagger Circle CI test failing again.

mayank408:
@the-dagger ^^

mayank408:
@the-dagger Resolved all the conflicts , now kindly merge this.

HimanshuS01:
Hey @mayank408 .I was talking about this.Web Search Result is not displaying.Help me to figure out the problem.Thanks
2bf19a4c-9e40-46f2-a7cd-d1ddb1cb1103

mayank408:
@HimanshuS01 Its not an issue , the web search take place only when the action is websearch , in this query the action is not websearch and the susi has the answer , so the answer is displayed . Thanks:)

HimanshuS01:
Thanks @mayank408 for helping me :) . But what kind of searches  will result into an action "websearch". Is there  any specific keyword to  write  query?

mayank408:
@HimanshuS01 No there is no specific keyword  , just if susi does not know the answer it will result in web search and as server kept updating so new answers also gets added.

HimanshuS01:
@mayank408  So I and you  are  getting different results for same query.i.e modi. So was it bcoz of server getting updated?

mayank408:
@HimanshuS01 Yup exactly I took these screenshots quite a few days ago , since then there must be update in the server. :)

PR #444, opened 2016-12-29
Fixes #443: Fixed the crashing of App and Added the myLocation Address

### LINKED ISSUE #443: App getting crashed after certain messages.
**Actual Behaviour**

Type: Where am I
Now the app getting crashed without showing the exact location.

**Expected Behaviour**

Type: Where am I
The response should show a marker on the map or tell the location of the user instead of showing the map only.

**Steps to reproduce it**

Ask Susi: Where am i
App will get crashed

Now re-run the app, there will be the response but without marker on the map.

**LogCat for the issue**

23307-23307/org.fossasia.susi.ai E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: org.fossasia.susi.ai, PID: 23307
                                                                      java.lang.NullPointerException: string arg == null
                                                                          at java.lang.String.concat(Native Method)
                                                                          at org.fossasia.susi.ai.activities.MainActivity$15$2.run(MainActivity.java:822)
                                                                          at android.os.Handler.handleCallback(Handler.java:739)
                                                                          at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                          at android.os.Looper.loop(Looper.java:148)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

**Screenshots of the issue**

Where-ever possible add a screenshot of the issue.

Yes I would like to work on this issue.

samsruti:
Fixes issue #443 : 
1. Fixed the crashing of app when someone asks "Where am I"
2. Added the "My location address" when someone asks "Where am I"

Screenshots for the change: 
bug_fixed

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -14,6 +14,8 @@ import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
+import android.location.Address;
+import android.location.Geocoder;
 import android.media.AudioManager;
 import android.net.ConnectivityManager;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -922,44 +948,47 @@ private void addNewMessage(String answer, boolean isMap, boolean isHavingLink, b
         } else {
             id = (long) temp + 1;
         }
-        if(answer.contains("google"))
-        {
+        if(answer!=null){
```
samsruti: @daminisatya Here I have added this condition so I provide some indentation down there

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -922,44 +948,47 @@ private void addNewMessage(String answer, boolean isMap, boolean isHavingLink, b
...
-        if(answer.contains("Calling")){
-            String splits[]=answer.split(" ");
-            startActivity(new Intent(Intent.ACTION_DIAL, Uri.fromParts("tel", splits[1], null)));
+            if(answer.contains("Calling")){
+                String splits[]=answer.split(" ");
+                startActivity(new Intent(Intent.ACTION_DIAL, Uri.fromParts("tel", splits[1], null)));
+            }
         }
 
+
```
mayank408: @samsruti I just don't get what have you done in above 30, 40 lines apart from indentation, pls avoid it , if a issue can be resolved in 20 lines there is no point in extending upto 60-70 lines. Thanks :)

samsruti: @mayank408 I think i explained the issue here: #443 Have a look.
And for your code that you implemented for Google Web Search, I would like to add something when the answer is null. If the answer will be null then, app will be getting crashed.
If you think
if(answer!=null){ ...} can avoid  indentation issues, then do suggest some ways.

mayank408: @samsruti I have implemented the google search only when the answer is not null , and also tested it thoroughly but had not encountered with any crash .It was working as expected . Thanks:)

samsruti: @mayank408 I was getting this error: java.lang.NullPointerException: string arg == null at that line if(answer.contains("google"))
So I thought of handling that error.
But now Its not happening. Will let you know if I encountered that again.
Thanks :)

PR #458, opened 2017-01-05
Fixes #455: Google Search Intent triggering in unintended cases resol… .

### LINKED ISSUE #455: Google Search Intent triggering in unintended cases
**Actual Behaviour**

The current feature to search google directly by writing "@google" has some issues. It is triggering the intent to search on google even in other cases when user did not ask to Google it.

**Expected Behaviour**

The Google intent must open only in required case i.e. only when user type "@google <query>"

**Steps to reproduce it**

Send a message "google.com"  to susi, the reply will open a Google search Chrome Custom Tabs intent which must not open. Reason is currently, the reply is being matched to contain "google" as keyword which should not be done.

**LogCat for the issue**

Not required

**Screenshots of the issue**

Not Required

Sure.

mayank408:
Fixes #455:Google Search Intent triggering in unintended cases resolved.


mayank408:
@daminisatya @the-dagger Now the google search will only take place when intended by the user.

PR #474, opened 2017-01-19
Fixes #471 Fixed typing dots

### LINKED ISSUE #471: Dots are showing before the message is sent.
**Actual Behaviour**
the dots are showing before the message is sent.
screenshot_20170115-004410

**Expected Behaviour**
The dots should shown after the message is delivered. As in any other IM apps. say ALLO.
and dots should get hide after reply.

**Steps to reproduce it**
@the-dagger is this a valid issue?

If its really an issue.

shubhamvashisht:
Fixes  #471 
Changes: Now the typing dots will be shown only after users message gets delivered to the server and hides after reply is set to the recyclerview.
however, dots are noticeable only if server takes some time give response back just like Allo.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -643,7 +643,6 @@ public class MainActivity extends AppCompatActivity {
         updateDatabase(id, query, DateTimeHelper.getDate(), false, true, false, false, false, false, isHavingLink, DateTimeHelper.getCurrentTime(), false, null);
         nonDeliveredMessages.add(new Pair(query, id));
         getLocationFromLocationService();
-        recyclerAdapter.showDots();
         new computeThread().start();
     }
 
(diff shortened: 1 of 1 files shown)


shubhamvashisht:
if an error occurs in susi_server, the response body will be empty?

Abhi2424shek:
The errors are shown like network no found so can you do that too, the dots ideally mean it's typing something {thinking}

daminisatya:
@shubhamvashisht Can you check the CircleCI fails?

shubhamvashisht:
@Abhi2424shek only error i have seen yet is "internet connectivity problem!" and it is occuring very often.
need to find if its because of server false response.

shubhamvashisht:
@daminisatya 
everything is fine. it says no view found matching (org.fossasia.susi.ai:id/down_angle), but its there.
how to build it again circleci?
gradleerrpr

shubhamvashisht:
@daminisatya mam, can you please try to build it again, Build on my local machine is not giving any such error.

Abhi2424shek:
It's passed so no worries.

jig08:
@shubhamvashisht Squash your commits into one and we are ready to go.

PR #500, opened 2017-02-01
Fixes #495 Now we can open/launch other apps from susi

### LINKED ISSUE #495: Now we can open/launch other apps from susi
**Expected Behaviour**

Other apps can be launched from susi by command/message open+APP NAME
issue

Yes, done it.

shubhamvashisht:
Fixes #495 Now we can open/launch other apps from susi

Screenshots for the change: 
g_20170201_2250343

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -7,6 +7,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.graphics.Bitmap;
@@ -90,6 +91,7 @@ import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Deque;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


shubhamvashisht:
@jig08 Please review this.

the-dagger:
What happens if there are more than 1 app containing the input string of AppName?

shubhamvashisht:
@the-dagger As a Hashmap cannot have duplicate keys, if two apps having same name are there then the latter one will replace first one. 
its very rare to have such conflict and even if there two apps with same name there functionality will be similar. for example if there are two camera apps with same label, any of that camera app will do the job.

the-dagger:
Ok.
What if the user asks susi to open messenger?
There is a Google Messenger and a Facebook Messenger
Ideally a picker should open up asking which app to open

shubhamvashisht:
Ok. will implement :)

Abhi2424shek:
Yep but make it beautifull @shubhamvashisht

jig08:
@shubhamvashisht Updates on this pls ?

shubhamvashisht:
I will send pr today.

Abhi2424shek:
@shubhamvashisht updates

shubhamvashisht:
@Abhi2424shek i am on it. I got stuck into something.

Abhi2424shek:
You are free to ask help @shubhamvashisht

mayank408:
@shubhamvashisht It will not work if we query like "Can you please open Facebook" . Therefore before adding any new features we have to add natural language processing and every action will come from the server not from the app itself.
 @the-dagger @jig08 Can we close this for now...

shubhamvashisht:
@mayank408  you are right.

PR #508, opened 2017-02-06
Fixes #507 : Toast issue resolved now toast is displayed only once

### LINKED ISSUE #507: Toast displayed multiple times
**Actual Behaviour**

"press again to exit" toast is displayed again after exiting the app.

**Expected Behaviour**

Toast should not be displayed the second time.

**Steps to reproduce it**

- open app
- press back 
- press back again

Yes

Jonsnow21:
Fixes issue #507 

Changes: if backPressedOnce is true, return after calling finish().


Abhi2424shek:
Can you check the failing tests and also provide some relevant screenshots like you did in the issue

Issue #522, opened 2017-02-13
mark messages as important


DravitLochan:
**Actual Behaviour**
As of now, a message can not be marked as important. Quite simple and straight. I had asked susi a question some time back. Now if I want to read the reply, either I ask susi again or scroll through the whole list of questions and answers to find it. 

**Expected Behaviour**
If we provide a feature to mark messages important, it will allow user to simply open the important one's and read through them.  This will allow the user to read the message even if there is no internet (solution to re-ask the question). In the absence of internet, this feature will be a bliss. 

**Steps to reproduce it**
Store the messages marked as important in the in-app database. 

**Would you like to work on the issue?**
Yes

DravitLochan:
1. show an option to mark as important onLongClick of a message.
2. when selected, give a toast message marked as important.
   in the background, store the message and its response in database.
3. add an option in overflow menu to open important messages.
4. list the messages from database.

is this workflow ok?

jig08:
@DravitLochan Sounds good.

DravitLochan:
I am using SQLite

mayank408:
@DravitLochan Use Realm instead as we are already using it and it is more secure and reliable than SQLite.

DravitLochan:
copy that!

betterclever:
@DravitLochan @mayank408 A better approach would be to add a starred field just like whatsapp in current model. When user clicks on important messages, filter can be applied.

DravitLochan:
Progress so far: on selecting one or more messages, * button comes along with delete, copy and share. On hitting *, selected messages are marked important (another attribute of isImportant has been added). Now I was thinking for  opening an activity which contains all the messages marked important. if some one has better solution, please tell.

DravitLochan:
may I close it?

Abhi2424shek:
Next time reference the issue solved in the commit message @DravitLochan. Close this one pls.

PR #548, opened 2017-02-20
Fixes # 522 : mark messages as important

DravitLochan:
Fixes issue #522 :mark messages as important

On selecting message(s), user gets an option to mark the message as important. on clicking it, the value of isImportant is set to true. 
Changes: an attribute of Important is added  to the ChatMessage.java. 
messages marked as important are put in log along with the total number of messages marked as important.
Screenshots for the change: 
screenshot_20170221-004055
screenshot_20170221-004224

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/ImportantMessages.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/ImportantMessages.java
@@ -1,0 +1,18 @@
+package org.fossasia.susi.ai.activities;
+
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.support.v7.widget.RecyclerView;
+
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/ImportantMessages.java
```diff
@@ -0,0 +1,18 @@
+package org.fossasia.susi.ai.activities;
+
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.support.v7.widget.RecyclerView;
```
DravitLochan: I have added RecyclerView to the activity which a user wants to open when he/she wants to find messages marked as important. the thing left is show the messages in the activity. will commit it.

PR #586, opened 2017-05-06
Fixes #565: Respond to a query as soon as connected to internet

### LINKED ISSUE #565: Respond to a query as soon as connected to internet
**Actual Behaviour**

If a user fires a query but is not connected to internet, They get the answer only when they open the app and the query gets fired on the server from the undeliveredMessage list. 

**Expected Behaviour**

Similar to other apps, Susi should be able to generate a notification with a response to the questions in the undeliveredMessage list.

**Steps to reproduce it**
Start a service which generates the notification as soon as it gets connected to the internet and answer the queries. 

**Let us know if this issue should be assigned to you or tell us who you think could help to solve this issue.**
yes!

DravitLochan:
fire notification on response

removed unwanted files

Fixes issue #[Add issue number here. Note: This will automatically close the issue. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Screenshots for the change:

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -61,7 +61,8 @@
             android:value="fabric_api_key" />
         -->
 
-        <activity android:name=".activities.ImportantMessages"></activity>
+        <activity android:name=".activities.ImportantMessages" />
+
     </application>
 
 </manifest>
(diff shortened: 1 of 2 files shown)


DravitLochan:
server is down so couldn't test. Will do and update

PR #596, opened 2017-05-08
Fixes #595: back button in Important messages activity

### LINKED ISSUE #595: back button in Important messages activity
**Actual Behaviour**

No such button

**Expected Behaviour**

A button on top left corner to switch back to main activity

**Steps to reproduce it**
add required method call

**Screenshots of the issue**

screenshot_20170508-214235

Yes

DravitLochan:
Fixes issue #595 

Changes: added a back button

Screenshots for the change: 
screenshot_20170508-214235

### DIFF
--- a/app/src/androidTest/java/org/fossasia/susi/ai/TestLoginActivity.java
+++ b/app/src/androidTest/java/org/fossasia/susi/ai/TestLoginActivity.java
@@ -7,6 +7,7 @@ import android.support.test.rule.ActivityTestRule;
 import android.support.test.runner.AndroidJUnit4;
 import android.util.Log;
 import android.view.WindowManager;
+import android.widget.RadioButton;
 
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/ImportantMessages.java
```diff
@@ -94,4 +98,16 @@ public void run() {
             }
         });
     }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem menuItem)
+    {
+        switch (menuItem.getItemId()) {
+            case android.R.id.home:
+                onBackPressed();
```
the-dagger: Not the best way to implement this imo.
It's better if you use the NavUtils class for providing up naivgation.
https://developer.android.com/training/implementing-navigation/ancestral.html#NavigateUp

PR #606, opened 2017-05-10
Fixes #605 : Clear all top activity before starting mainactivity

### LINKED ISSUE #605: Clear all  top activity before starting mainactivity
**Actual Behaviour**

There are some activity which remain in back stack after mainactivity is running. So when user press backbutton instead of closing app those activity comes on top.

**Expected Behaviour**

It should not happen.

**Steps to reproduce it**

Click on forgot password button and in forgot password activity click on reset button  after that login.
**LogCat for the issue**

**Screenshots of the issue**

yes

amitiwary999:
Fixes issue #605 

Changes: 

Screenshots for the change:


app/src/main/java/org/fossasia/susi/ai/activities/ForgotPasswordActivity.java
```diff
@@ -136,6 +136,7 @@ public void onResponse(Call<ForgotPasswordResponse> call, Response<ForgotPasswor
                                 public void onClick(DialogInterface dialog, int id) {
                                     Intent intent = new Intent(ForgotPasswordActivity.this, LoginActivity.class);
                                     startActivity(intent);
+                                    finish();
```
chiragw15: @amitiwary999 When login activity is started from forgot password activity, we have to press back twice to exit app. Pressing back once opens previous instance of login activity. Either use intent flags `intent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK | Intent.FLAG_ACTIVITY_NEW_TASK);` or override onBackPressed method in login activity and call finishAffinity().

PR #608, opened 2017-05-11
Fixes #96 : Added support for tables

### LINKED ISSUE #96: Add support for tables in Susi response
Susi server provides us data in the form of a list. The response should also focus on displaying such data along with the response. It must show it in the form of a bulleted list. Susi shows the type as table in its response

chiragw15:
Fixes issue #96 

Changes: Added support for tables.

Screenshots for the change: 
screenshot_20170511-074742
screenshot_20170511-075914

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -853,7 +853,6 @@ public class MainActivity extends AppCompatActivity {
                                                  .get(0).getExpression();
                                         List<String> urlList = extractUrls(answer);
                                         Log.d(TAG, urlList.toString());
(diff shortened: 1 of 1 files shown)


betterclever:
@chiragw15 The support for tables is not meant like this imo. There can be a scrollable cards. We can make a ViewHolder specifically for this. This looks cluttered. 

@Orbiter @the-dagger Please guide how should it be done.

chiragw15:
@betterclever it is cluttered. I implemented it similar to http://susi.ai/chat . I don't know if scrollable cards will look good since we don't have links for articles right now. Plus some of the titles are only 1 word whereas some are more than 3-4 lines. Need to think something else.

betterclever:
@chiragw15 Yes, it needs some research.

PR #617, opened 2017-05-15
Removing dependency of ChatMessageView

### LINKED ISSUE #611: Removing dependency on ChatmessageView
**Actual Behaviour**

We have this `me.himanshusoni.chatmessageview:chat-message-view:1.0.3` dependency in gradle file which is used only in [item_susi_image.xml](https://github.com/fossasia/susi_android/blob/development/app/src/main/res/layout/item_susi_image.xml) , [item_user_image.xml](https://github.com/fossasia/susi_android/blob/development/app/src/main/res/layout/item_user_image.xml) and [search_list.xml](https://github.com/fossasia/susi_android/blob/development/app/src/main/res/layout/search_list.xml)

**Expected Behaviour**

We can replace Chat Message view with card view as previously done here. https://github.com/fossasia/susi_android/pull/503 . 
@betterclever was there a specific reason for leaving these three files ?

yes

chiragw15:
Fixes issue #611 

Changes: Removed dependency of ChatMessageVIew by replacing it with CardView.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -52,7 +52,6 @@ dependencies {
     compile "com.android.support:customtabs:$support_lib_version"
     compile "com.android.support:cardview-v7:$support_lib_version"
 
-    compile 'me.himanshusoni.chatmessageview:chat-message-view:1.0.3'
     compile 'com.github.PhilJay:MPAndroidChart:v2.2.5'
 
     /*
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/viewholders/SearchResultsHolder.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/viewholders/SearchResultsHolder.java
@@ -16,7 +16,7 @@ import butterknife.ButterKnife;
 
 public class SearchResultsHolder extends RecyclerView.ViewHolder {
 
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.2.3'
```
betterclever: Don't downgrade grade logon version

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
```
betterclever: Here too

betterclever:
LGTM

PR #620, opened 2017-05-16
Fixes #619 : Voice output for voice input pref fixed

### LINKED ISSUE #619: Bug in "Speech output for speech input" pref
**Actual Behaviour**

Speech output doesn't work for first time but works if we disable and then enable it again. That's an issue. There is something wrong with dealing of prefs.

**Expected Behaviour**

It should work for every case.

yes

chiragw15:
Fixes issue #619 

Changes: In the first run after installing app, the voice output was not working. Fixed it.


mariobehling:
No need to add "fixes xyz" to commit message. Please avoid and focus the commit message on the actual changes made.

PR #626, opened 2017-05-19
Fixes #624 : Leak Canary enabled

### LINKED ISSUE #624: Enabling leak canary for leak detection
**Actual Behaviour**

There are memory leaks but they aren't detected.

**Expected Behaviour**

Enabling leak canary to detect memory leaks.

yes

chiragw15:
Fixes issue #624 
Changes: Added leak canary for memory leak detection

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/MainApplication.java
+++ b/app/src/main/java/org/fossasia/susi/ai/MainApplication.java
@@ -2,6 +2,8 @@ package org.fossasia.susi.ai;
 
 import android.app.Application;
 
+import com.squareup.leakcanary.LeakCanary;
+
 import io.realm.Realm;
 import io.realm.RealmConfiguration;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/susi/ai/MainApplication.java
```diff
@@ -23,6 +25,14 @@ public static MainApplication getInstance() {
     @Override
     public void onCreate() {
         super.onCreate();
+
```
amitiwary999: why extra space here?

app/src/main/java/org/fossasia/susi/ai/MainApplication.java
```diff
@@ -23,6 +25,14 @@ public static MainApplication getInstance() {
...
+        if (LeakCanary.isInAnalyzerProcess(this)) {
+            // This process is dedicated to LeakCanary for heap analysis.
+            // You should not init your app in this process.
+            return;
+        }
+        LeakCanary.install(this);
+
```
amitiwary999: same here

app/src/main/java/org/fossasia/susi/ai/MainApplication.java
chiragw15: @amitiwary999 These 2 empty lines (1 in the starting and 1 at the end) are there to separate leak canary code from other code. This is to prevent accidental deletion of other useful code when we'll remove the leak canary code in future.

betterclever: @amitiwary999 It is for legibility only as far as I see. Why do you want to remove these lines?

amitiwary999: I don't want to remove these lines .I am just asking why extra space is used here.

PR #634, opened 2017-05-25
Fixes #633 : Adding feature for retrieving old chat messages

### LINKED ISSUE #633: Digging out previous chat messages at time of user login
**Actual Behaviour**
Currently, if a user logins, their previous messages are not displayed. 

**Expected Behaviour**
At every login, fetch the old chat messages and display them.

**Steps to reproduce it**
We have a new memory servlet which will help. The previous conversation log is returned in a cognitions object.

chiragw15:
Fixes issue #633
Changes: Adding feature for retrieving old chat messages using memory.json
Used a dialog box to confirm from user if she wants to retrieve her messages.

Screenshots for the change: 
screenshot_20170525-230651
screenshot_20170525-230654 1

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -1,6 +1,7 @@
 package org.fossasia.susi.ai.activities;
 
 import android.Manifest;
+import android.app.ProgressDialog;
 import android.content.ActivityNotFoundException;
 import android.content.BroadcastReceiver;
 import android.content.Context;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -157,6 +159,7 @@
     private BroadcastReceiver networkStateReceiver;
     private ClientBuilder clientBuilder;
     private Deque<Pair<String, Long>> nonDeliveredMessages = new LinkedList<>();
+    ProgressDialog progressDialog;
```
mayank408: @chiragw15 You can use private here.

mariobehling:
Please avoid question to retrieve old messages. Simply retrieve old messages. If number of messages is too big retrieve a limited number of old messages and more old messages when the user scrolls back. Please open an issue for that.

Issue #635, opened 2017-05-27
Memory Leak in Location Helper


chiragw15:
**Actual Behaviour**

There is a memory leak in location helper.

**Expected Behaviour**

There should be no memory leak in the app.

**Steps to reproduce it**

Be sure your location is ON . Then start the app and query "where am I" and then close the app. You will see memory leak.

**Screenshots of the issue**

![screenshot_20170527-144157](https://cloud.githubusercontent.com/assets/16978820/26522032/0779ec4c-4315-11e7-84dc-6e9cc3298bba.png)

**Would you like to work on the issue?**
yes

chiragw15:
@iamareebjamal I used `locationManager.removeUpdates(this);` in the onStop() method but it didn't fix the issue. I googled it and found that is an issue in google maps sdk as mentioned in this [PR](https://github.com/mcharmas/Android-ReactiveLocation/pull/142). There is a work around using weak reference [here](https://github.com/googlesamples/android-play-location/issues/26#issuecomment-248101802). Need help in implementing it.

iamareebjamal:
Sure, first create a topic branch in your fork and I can help with that.

chiragw15:
@iamareebjamal https://github.com/chiragw15/susi_android/tree/locationLeak this the branch

iamareebjamal:
Hey, I have commented on your commit in the branch. Please take a look if issue is not solved yet

PR #636, opened 2017-05-27
[Do not merge] fix : memory leak in location helper 

chiragw15:
Fixes issue #635 

Changes: Fixed memory leak in location helper. Removed location updates in onStop() method.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -135,7 +135,7 @@ public class MainActivity extends AppCompatActivity {
     protected DotsTextView voiceDots;
     @BindView(R.id.cancel)
     protected ImageView cancelInput;
-    
+
     private boolean atHome = true;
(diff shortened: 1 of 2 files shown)


chiragw15:
@amitiwary999 I was going to remove the listener in onDestroy() but since it is not guaranteed that the onDestroy() method will be called, (ref [link](https://stackoverflow.com/questions/18361719/android-activity-ondestroy-is-not-always-called-and-if-called-only-part-of-the) ) I did that in onStop(). So, should initiate the locationHelper in onStart()? What difference does it make? Can you give me a link or something which explains this?

mariobehling:
Please avoid "Fixes 365" in the commit message. Only add the issue into the description.

amitiwary999:
@chiragw15 according to lifecycle of android on pressing home button onStop() is called and then again on starting app onStart() method will be called first.[https://stackoverflow.com/questions/5019686/what-methods-are-invoked-in-the-activity-lifecycle-in-the-following-cases](link). Same when we move from one activity to other.

Issue #637, opened 2017-05-29
Removing dialog box asking user to retrieve old messages


chiragw15:
**Actual Behaviour**

There is a dialog box which asks user if he wants to retrieve old messages

**Expected Behaviour**

As mentioned here https://github.com/fossasia/susi_android/pull/634#issuecomment-304710323, removing the dialog box asking user to retrieve old messages and loading previous messages automatically.

**Would you like to work on the issue?**
@mayank408 's working

mayank408:
@chiragw15 Can I work on this.

chiragw15:
Sure, go ahead :+1:

PR #640, opened 2017-05-30
Dialog box is removed from getting old messages.

mayank408:
Dialog box is removed from getting old messages. Now previous messages get automatically loaded in the chat.

Fixes issue #637

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -292,8 +292,7 @@ public class MainActivity extends AppCompatActivity {
 
     private void getOldMessages() {
         if (isNetworkConnected()) {
-            ClientBuilder clientBuilder1 = new ClientBuilder();
-            Call<MemoryResponse> call = clientBuilder1.getSusiApi().getChatHistory();
+            Call<MemoryResponse> call = clientBuilder.getSusiApi().getChatHistory();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -328,7 +328,8 @@ private void parseSusiResponse(SusiResponse susiResponse) {
 
     private void getOldMessages() {
         if (isNetworkConnected()) {
-            Call<MemoryResponse> call = clientBuilder.getSusiApi().getChatHistory();
+            ClientBuilder clientBuilder1 = new ClientBuilder();
+            Call<MemoryResponse> call = clientBuilder1.getSusiApi().getChatHistory();
```
betterclever: Use the clientBuilder created in onCreate here, no need to initialize a new object.

mayank408: Yes I was previously using that but it could not be initialized and getaoldMessage gets called before it. Thus it was giving null pointer exception.

betterclever: Exception was there since you called retrieveOldMessage before initializing clientBuilder. Initialize it above call to retrieveOldMessages() in onCreate ()

mayank408: Changes done

Issue #639, opened 2017-05-29
Retrieving old messages not shown correctly  in case of websearch


amitiwary999:
**Actual Behaviour**

Retrieving old messages is not shown correctly.

**Expected Behaviour**

Retrieved messages should be shown correctly.

**Steps to reproduce it**

Start chat with susi which also include websearch. Now logout and again login. you can see websearch is shown according to latest websearch.

**LogCat for the issue**

**Screenshots of the issue**

![screenshot_2017-05-30-02-45-13](https://cloud.githubusercontent.com/assets/14003244/26562359/4fc2b0ea-44e3-11e7-9163-a801ff682d7f.png)
![screenshot_2017-05-30-02-45-46](https://cloud.githubusercontent.com/assets/14003244/26562360/4fc7a4ec-44e3-11e7-9be9-c2d339a89e91.png)

**Would you like to work on the issue?**

yes

chiragw15:
This line `webquery = MainActivity.webSearch;` [here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L486) is causing trouble. If the webSearch string is changed too fast i.e before loading of view, the results will be displayed of the last query. Change the implementation of websearch.

PR #642, opened 2017-05-31
Fix:Retrieved old messages show wrong result in case of websearch

amitiwary999:
Fixes issue #639 

Changes: 

Screenshots for the change:


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -321,6 +321,7 @@ private void parseSusiResponse(SusiResponse susiResponse) {
             isWebSearch = susiResponse.getAnswers().get(0).getActions().get(1).getType().equals("websearch");
             datumList = susiResponse.getAnswers().get(0).getData();
             webSearch = susiResponse.getAnswers().get(0).getActions().get(1).getQuery();
+            answer = answer+"#"+webSearch;
```
mayank408: Could you pls explain what are you trying to achieve here. Thanks

amitiwary999: Currently when we retrieve old messages then variable webSearch contains only one only which is last retrieved websearch. So [webquery](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L486)  can have only that single value. Thats why retrieved messages show same webresult. Here I attached each websearch with answer so that when recyclerview inflate with text answer [here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L483), it will show corresponding websearch.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -483,18 +483,17 @@ public boolean onLongClick(View view) {
     private void handleItemEvents(final WebSearchHolder websearchholder, final int position)  {
         websearchholder.backgroundLayout.setBackgroundColor(ContextCompat.getColor(currContext, isSelected(position) ? R.color.translucent_blue : android.R.color.transparent));
 
-        webquery = MainActivity.webSearch;
-
         final ChatMessage model = getData().get(position);
+        int checkIndex = model.getContent().lastIndexOf("#");
         websearchholder.messageStar.setVisibility( (model.isImportant()) ? View.VISIBLE : View.GONE);
-        websearchholder.text.setText(model.getContent());
+        websearchholder.text.setText(model.getContent().substring(0,checkIndex));
```
chiragw15: What if webquery has # in it?

amitiwary999: We are showing websearch when we are not getting reply from susi and that case text is simply I don't know how to answer this here is the websearch result: . There is no scope for #. Can you provide any ?

chiragw15: if you query something like "search for dog", the server responses with 
answer : Here is a search result
websearch : dog
What you are doing is you are converting `answer` to `answer+"#"+websearch`
 So in the above case answer will become `Here is a search result#dog` 
Then you find index of "#" using `int checkIndex = model.getContent().lastIndexOf("#");` and separate them again.
Now suppose I query "search for abc#abc"
answer : Here is a search result
websearch : abc#abc
Now when you combine them answer will become `Here is a search result#abc#abc`.
Now when you use `int checkIndex = model.getContent().lastIndexOf("#");` and separate them, you'll get answer as `Here is a search result#abc` and websearch as `abc`.
</br>
Though I am not getting expected result "search for abc#abc". The server is interpreting '#' as space and ignoring it. Need to confirm it from @Orbiter

chiragw15: Just confirmed with orbiter, the # will be interpreted as whitespace. So, most probably there won't be '#' in result. So, it may work. But, I am not that happy with overall implementation of websearch result. I am creating a new issue for improving it. This can be merged for now. Just resolve the conflicts.

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
iamareebjamal: Please use `StringBuilder` instead of String concatenation

mayank408:
@amitiwary999 pls squash your commits.

chiragw15:
Fixed in #659

PR #644, opened 2017-05-31
Dialog box is removed from asking the user for getting old messages.

mayank408:
Fixes issue #637 
Dialog box is removed from asking the user for getting old messages. Now the previous messages will be automatically loaded in the chat.


mayank408:
@mariobehling Pls merge this.

Issue #651, opened 2017-06-02
Issue with alignment of layout in display of maps.


mayank408:
**Actual Behaviour**

The layout is not uniform.

**Expected Behaviour**

The layout must be uniform.

**Screenshots of the issue**

![screenshot_1496417839](https://cloud.githubusercontent.com/assets/22056321/26733179/a6004886-47d7-11e7-8233-60070a145ab2.png)

**Would you like to work on the issue?**

Yes

chiragw15:
Map is now a separate message.(After #681) Thus no need for it now.

PR #653, opened 2017-06-03
Fix UI issues with the map view in the chat.

mayank408:
Fixes issue #651 
Alignment of text view is corrected.

Screenshots for the change: 

screenshot_1496493749


app/src/main/res/layout/item_susi_map.xml
```diff
@@ -30,7 +30,7 @@
 
                 <TextView
                     android:id="@+id/text"
-                    android:layout_width="wrap_content"
+                    android:layout_width="240dp"
```
iamareebjamal: Hardcoding width is not the best idea. On tablets this will look really small

iamareebjamal: The ultimate issue is with gravity, right?

amitiwary999: I think issue is text s shown at right side instead of left side.

amitiwary999: @iamareebjamal So yes issue is with gravity of text.

iamareebjamal: Yeah, that's about gravity of the text. Changing it to constant is not the best way to solve this. Try setting `layout_gravity` on the container layout or `gravity` on the text itself

mayank408: The Problem is we are using card view of fixed size ie 250dp

iamareebjamal: Then `match_parent` should do it. The problem with hardcoding is that if we change one of the dimens, we will have to change all and that will be difficult to track

mayank408: @iamareebjamal We are using custom layout for this, therefore `match_parent`is not working as excepted. Could you help me out in this.

iamareebjamal: Sure, I'll pull your changes and have a look

Issue #657, opened 2017-06-03
Crash on cancelling speech input


iamareebjamal:
**Actual Behaviour**

App crashes if speech input is enabled and then disabled

**Expected Behaviour**

The app should not crash

**Steps to reproduce it**

Go to chat screen, click on the microphone icon to enable speech input, then click on cross icon, the app crashes

**LogCat for the issue**

````
java.lang.IllegalArgumentException: Service not registered: android.speech.SpeechRecognizer$Connection@fb3c905
  at android.app.LoadedApk.forgetServiceDispatcher(LoadedApk.java:1289)
  at android.app.ContextImpl.unbindService(ContextImpl.java:1511)
  at android.content.ContextWrapper.unbindService(ContextWrapper.java:648)
  at android.speech.SpeechRecognizer.destroy(SpeechRecognizer.java:408)
  at org.fossasia.susi.ai.activities.MainActivity.cancelSpeechInput(MainActivity.java:257)
  at org.fossasia.susi.ai.activities.MainActivity_ViewBinding$1.doClick(MainActivity_ViewBinding.java:44)
  at butterknife.internal.DebouncingOnClickListener.onClick(DebouncingOnClickListener.java:22)
  at android.view.View.performClick(View.java:5609)
  at android.view.View$PerformClick.run(View.java:22259)
  at android.os.Handler.handleCallback(Handler.java:751)
  at android.os.Handler.dispatchMessage(Handler.java:95)
  at android.os.Looper.loop(Looper.java:154)
  at android.app.ActivityThread.main(ActivityThread.java:6077)
  at java.lang.reflect.Method.invoke(Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)

````

**Would you like to work on the issue?**

No

amitiwary999:
@iamareebjamal I tried it but app is not crashing in my case.can you tell me exact step?

iamareebjamal:
![](https://im3.ezgif.com/tmp/ezgif-3-32d8ad45f5.gif)

iamareebjamal:
Happening continously

chiragw15:
Though I was not able to reproduce the crash but I found something similar on stackoverflow https://stackoverflow.com/questions/13227105/speechrecognizer-not-connected-to-recognition-service.
#661 should prevent the crash.

[...]

PR #661, opened 2017-06-04
fix: Fixes crash when voice input is closed

chiragw15:
Fixes issue #657 

Changes: Added a check to see if recognizer is null.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -66,6 +66,7 @@ import org.fossasia.susi.ai.R;
 import org.fossasia.susi.ai.adapters.recycleradapters.ChatFeedRecyclerAdapter;
 import org.fossasia.susi.ai.helper.Constant;
 import org.fossasia.susi.ai.helper.DateTimeHelper;
+import org.fossasia.susi.ai.helper.MediaUtil;
 import org.fossasia.susi.ai.helper.PrefManager;
 import org.fossasia.susi.ai.model.ChatMessage;
 import org.fossasia.susi.ai.rest.clients.BaseUrl;
@@ -236,8 +237,11 @@ public class MainActivity extends AppCompatActivity {
 
     @OnClick(R.id.cancel)
     public void cancelSpeechInput() {
-        recognizer.cancel();
-        recognizer.destroy();
+        if(recognizer != null) {
+            recognizer.cancel();
+            recognizer.destroy();
+            recognizer=null;
+        }
         hideVoiceInput();
     }
 
@@ -257,6 +261,10 @@ public class MainActivity extends AppCompatActivity {
 
     public static boolean checkSpeechAlwaysPref() {
         return PrefManager.getBoolean(Constant.SPEECH_ALWAYS, false);
+    }
+
+    public boolean checkMicInput() {
+        return micCheck = MediaUtil.isAvailableForVoiceInput(MainActivity.this);
     }
 
     private void parseSusiResponse(SusiResponse susiResponse) {
@@ -405,8 +413,7 @@ public class MainActivity extends AppCompatActivity {
 
         if(ActivityCompat.checkSelfPermission(this,
                 Manifest.permission.RECORD_AUDIO) == PackageManager.PERMISSION_GRANTED) {
-            micCheck = true;
-            PrefManager.putBoolean(Constant.MIC_INPUT, true);
+            PrefManager.putBoolean(Constant.MIC_INPUT, checkMicInput());
         }
 
         getLocationFromLocationService();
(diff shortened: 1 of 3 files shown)


iamareebjamal:
I'll pull and see if this fixed the issue

iamareebjamal:
This did not fix it. But I digged into the error. The problem was that my emulator didn't have speech input system, so the SpeechRecognizer was logging error that
```
E/SpeechRecognizer: no selected voice recognition service
```

So, I opened the class and saw that it also notifies the listener with `onError` callback : 
````Java
if (mServiceComponent == null) {
    String serviceComponent = Settings.Secure.getString(mContext.getContentResolver(),
            Settings.Secure.VOICE_RECOGNITION_SERVICE);
    
    if (TextUtils.isEmpty(serviceComponent)) {
        Log.e(TAG, "no selected voice recognition service");
        mListener.onError(ERROR_CLIENT);
        return;
    }
    
    serviceIntent.setComponent(ComponentName.unflattenFromString(serviceComponent));                
} else {
    serviceIntent.setComponent(mServiceComponent);
}
````

But the listener isn't `RecognitionListener`, it is some internal listener which then sends the message to Recognition listener. But I checked and onError is never called for some reason.

I also noticed that there is a GC call right after I press the mic option, so if they are using WeakReference, then listener can be dereferenced at that time and become null.

I tried thinking of a solution but when the callbacks aren't working, there is nothing much we can do except to pull the internal code outside to check if there are no speech input methods and not try to start listening if this is the case

**Comment by chiragw15** [comment]
So, what do you suggest? I guess, this problem of not having a speech input method should not occur on a mobile device. I had tested this for 2-3 mobile phones, worked fine for them. So, what should we do? `except to pull the internal code outside to check if there are no speech input methods and not try to start listening if this is the case` Should we implement this?

**Comment by iamareebjamal** [comment]
The thing is, application should not crash in any condition. A simple check won't hurt

**Comment by chiragw15** [comment]
@iamareebjamal can you tell me which emulator you are using? I have done some changes. Needed to test them.

**Comment by iamareebjamal** [comment]
Genymotion

**Comment by chiragw15** [comment]
@iamareebjamal review please.

File: app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java [hunk 1]
```diff
@@ -405,8 +409,13 @@ protected void onCreate(Bundle savedInstanceState) {
 
         if(ActivityCompat.checkSelfPermission(this,
                 Manifest.permission.RECORD_AUDIO) == PackageManager.PERMISSION_GRANTED) {
-            micCheck = true;
-            PrefManager.putBoolean(Constant.MIC_INPUT, true);
+            if(MediaUtil.isAvailableForVoiceInput(MainActivity.this)) {
```
iamareebjamal: Store in a boolean `voiceInputAvailable` and 

````Java
micCheck = voiceInputAvailable;
PrefManager.putBoolean(Constant.MIC_INPUT, voiceInputAvailable);
````

File: app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java [hunk 2]
```diff
@@ -726,8 +735,13 @@ public void onRequestPermissionsResult(int requestCode, String permissions[], in
                     micCheck = false;
                     PrefManager.putBoolean(Constant.MIC_INPUT, false);
                 } else {
-                    micCheck = true;
-                    PrefManager.putBoolean(Constant.MIC_INPUT, true);
+                    if(MediaUtil.isAvailableForVoiceInput(MainActivity.this)) {
```
iamareebjamal: Same here

app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -752,8 +766,13 @@ public void onRequestPermissionsResult(int requestCode, String permissions[], in
...
```
iamareebjamal: Same here, why not create a function?

app/src/main/java/org/fossasia/susi/ai/activities/SettingsActivity.java
```diff
@@ -84,6 +89,20 @@ public void execute(Realm realm) {
                     return true;
                 }
             });
+
+            micSettings = getPreferenceManager().findPreference("Mic_input");
+            if(ActivityCompat.checkSelfPermission(getContext(),
+                    Manifest.permission.RECORD_AUDIO) == PackageManager.PERMISSION_GRANTED) {
+                if(MediaUtil.isAvailableForVoiceInput(getContext())) {
```
iamareebjamal: Same here

app/src/main/java/org/fossasia/susi/ai/helper/MediaUtil.java
```diff
@@ -0,0 +1,54 @@
...
+        MediaRecorder recorder = new MediaRecorder();
+        recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+        recorder.setOutputFormat(MediaRecorder.OutputFormat.DEFAULT);
+        recorder.setAudioEncoder(MediaRecorder.AudioEncoder.DEFAULT);
+        recorder.setOutputFile(new File(context.getCacheDir(), "MediaUtil#micAvailTestFile").getAbsolutePath());
+        boolean available = true;
+        try {
+            recorder.prepare();
+        }
+        catch (IOException exception) {
```
iamareebjamal: Move catch up

chiragw15:
@iamareebjamal done.

Issue #658, opened 2017-06-03
Crash on query "Play some Music"


iamareebjamal:
**Actual Behaviour**

Although I can not reproduce this bug, but the first time I asked to "Play some Music", the app crashed

**Expected Behaviour**

App should not crash

**Steps to reproduce it**

Ask susi to "Play some Music"

**LogCat for the issue**

````
Process: org.fossasia.susi.ai, PID: 21578
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
    at org.fossasia.susi.ai.activities.MainActivity.voiceReply(MainActivity.java:778)
    at org.fossasia.susi.ai.activities.MainActivity.access$2800(MainActivity.java:116)
    at org.fossasia.susi.ai.activities.MainActivity$18.onResponse(MainActivity.java:1188)
    at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:68)
    at android.os.Handler.handleCallback(Handler.java:751)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:154)
    at android.app.ActivityThread.main(ActivityThread.java:6077)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
````

**Would you like to work on the issue?**

No

amitiwary999:
@iamareebjamal it is not related to query. I too get this error only  once when I opened app first time and after that I didn't get this error. I am trying to reproduce this error.

chiragw15:
https://circleci.com/gh/chiragw15/susi_android/15 Found this error in Circle CI build also

chiragw15:
https://github.com/fossasia/susi_android/pull/659/commits/907ed24167ad9011a16e34d7ef79a55d8fd93414 Does doing this fixes the above issue? Because by doing this, Circle CI build passed. @iamareebjamal @amitiwary999

iamareebjamal:
Umm... Maybe. But as we cannot confirm, I'll say you should convert all class types to primitives in a PR so that there are no future problems

PR #662, opened 2017-06-04
fix : Changed all Boolean to boolean 

chiragw15:
Fixes issue #658 

Changes: Changed Boolean to boolean

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -147,9 +147,9 @@ public class MainActivity extends AppCompatActivity {
     private boolean isSearchResult;
     private boolean isWebSearch;
     private List<Datum> datumList = null;
-    private Boolean micCheck;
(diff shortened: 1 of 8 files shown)


iamareebjamal:
@chiragw15 Wouldn't it be better if all class types are changed to primitives in a single PR?

chiragw15:
I did that only. These are the ones I found. Did I miss some of them?

iamareebjamal:
I was talking about all class types, not just boolean, over the whole project

chiragw15:
Alright, I thought you meant all Boolean over the project. Thanks. Will do the rest too.

chiragw15:
@iamareebjamal Should the Class types in POJO class of retrofit responses be changed to primitive type too? Example [here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/Datum.java#L24)

iamareebjamal:
Yes, they should be

chiragw15:
@iamareebjamal done. Please review.

PR #675, opened 2017-06-07
Fix UI issues with the map view in the chat

mayank408:
Fixes issue #651 

Screenshots for the change: 

screenshot_1496493749


app/src/main/res/values/dimens.xml
```diff
@@ -29,5 +29,6 @@
     <dimen name="margin_very_small">2dp</dimen>
     <dimen name="margin_large">10dp</dimen>
     <dimen name="message_textview_bottom_padding">3dp</dimen>
+    <dimen name="textview_layout_max_width">240dp</dimen>
```
chiragw15: It was suggested by @iamareebjamal in #653 that fixing value is not a solution here.

mayank408: @chiragw15 The problem is that we are using custom views because of this match_parent is not working as expected as suggested by @iamareebjamal. I have tried tweaking the custom view as well but could not find a more probable solution for this.

chiragw15: One sec, sorry, you can close this PR since after adding loop to display result of all action types independently as a new message, this issue will be solved. Pr #681

Issue #688, opened 2017-06-09
Current html parsing in susi_android is not implemented correctly


amitiwary999:
**Actual Behaviour**

If html tag contain href then it should be converted to link.

**Expected Behaviour**

Currently after parsing html tag are converted to string  that's why href tag is not parsed properly.

**Would you like to work on the issue?**

yes

chiragw15:
I guess I fixed it in #681 . See removed some code from [here](https://github.com/fossasia/susi_android/pull/681/commits/dabc98602d989fd7e85ed0e7c8758c734324426d#diff-97eb5665ef5431161882b9bd50d1843dL1086), added some code [here](https://github.com/fossasia/susi_android/pull/681/commits/dabc98602d989fd7e85ed0e7c8758c734324426d#diff-f14e5d3cf86ea64cf8038542ccaa4185R445) and [here](https://github.com/fossasia/susi_android/pull/681/commits/dabc98602d989fd7e85ed0e7c8758c734324426d#diff-b68ec3c378db0ddf241a4cc542ecf9e3R35). Is that what you are talking about? It worked for me.

amitiwary999:
Yes but you did not mention that you will work on that. I asked you yesterday about it but didn't reply. Please let me also solve some issue. If you solve all the issue what I will do.

chiragw15:
I am really sorry about this but I did reply you ([here](https://github.com/fossasia/susi_android/pull/683#discussion_r120833402))that I will look into it and let you know. I google searched for this problem and wasted about 2 hours on this and finally got the solution by hit and trial. I had no intention as such to solve this issue but you asked for help, that's why I tried implementing it and it worked. I think it is not fully implemented also. Go ahead and send a PR with the changes.

chiragw15:
main catch was to remove .toString() when parsing.

amitiwary999:
I too wasted 1 hour yesterday for checking same thing

amitiwary999:
then I found that it must be implemented in recycleradapter for better result. thanks  I will send a pull request

chiragw15:
No worries, go ahead and send the PR for changes. Cheers. :+1:

PR #689, opened 2017-06-09
Fix : Current html parsing in susi_android is not implemented correctly

amitiwary999:
Fixes issue #688 

Changes: .toString was creating problem so removed it and uses parsing in recycleradapter instead of mainactivity.


chiragw15:
@amitiwary999 Please update the branch and add code in the place of TODO. Thanks :+1:

PR #684, opened 2017-06-08
Fix : Reduce Codacy issues wherever possible

amitiwary999:
Fixes issue #682 

Changes: some changes as mentioned [here](https://www.codacy.com/app/harshithdwivedi/susi_android/issues?step=0&filters=W3siaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6WyJFcnJvciBQcm9uZSJdfV0%3D)

PR #690, opened 2017-06-09
Adding alert dialog box to a single helper class

mayank408:
Adding dialog box in the helper class to remove redundancy.

@chiragw15 Can we implement the alert function way. Pls give feedback. I will be working on finding the solution for button callbacks.


chiragw15:
Isn't this and #684 same?

mayank408:
@chiragw15 I am looking for button callbacks which is different. I don't know exactly how can we achieve this.

Issue #685, opened 2017-06-08
Update "rss" action type


amitiwary999:
**Expected Behaviour**

Recently rss action type is updated in susi_server. So we have to implement it in client side.

**Would you like to work on the issue?**

yes

mayank408:
i@amitiwary999 @chiragw15 I want to work on this as my issue related to map UI got resolved .

chiragw15:
@amitiwary999 @mayank408 rss action type is already implemented and working. Just check if it needs some tweaks or changes according to the new response. And improve performance.
![screenshot_20170609-045716](https://user-images.githubusercontent.com/16978820/26954909-c8b61e70-4cd0-11e7-8be3-478d3ad5fb27.png)

PR #693, opened 2017-06-10
Fix : update rss action type [Work in progress]

amitiwary999:
Fixes issue #685 

Changes:


amitiwary999:
in websearch we are showing result after parsing result from duckduckgo . But in rss we have to show only description title and link which we get from susi reply. No extra thing like image from websearch etc. that's why I use different  model for it. I will try to check if it will be better to use it along with searhresultholder

amitiwary999:
after mergin new updates my last commit removed automatically. my new pr is
[here](https://github.com/fossasia/susi_android/pull/694)

Issue #697, opened 2017-06-14
issue with message history


mayank408:
**Actual Behaviour**

The retrieved messages from Susi memory are shown with the current date in the app.

**Expected Behaviour**

The retrieved messages from Susi memory should be shown with their actual date in the app.

**Steps to reproduce it**

Change the date of our phone and the date of retrieved messages will also change.

**Would you like to work on the issue?**
Yes

PR #704, opened 2017-06-15
Fix issue with date of retrieved messages.

mayank408:
Fixes issue #697 
Now the date of the retrieved messages is set according to the query date coming from the response from the server.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -82,6 +82,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -1070,20 +1091,20 @@ public void onFailure(Call<SusiResponse> call, Throwable t) {
         }
     }
 
-    private void updateDatabase(final long prevId, final String message, final boolean isDate, final boolean mine,
+    private void updateDatabase(final long prevId, final String message, final boolean isDate, final String Date, final boolean mine,
```
betterclever: Please strictly follow variable naming convention. Variable name starts with lowercase.

mayank408:
@betterclever The timezone is local.

betterclever:
@mayank408 So, server sends date and time in local timezone. Right?

mayank408:
@betterclever We send timezone at the time of query which is saved in server. So yes It is local

betterclever:
@mayank408 Okay.

Issue #717, opened 2017-06-16
Remove Important message activity


chiragw15:
**Actual Behaviour**

There is a feature in app to mark messages as important, it does not involve server and stores important messages locally.

**Expected Behaviour**

As stated by Mario, since the server doesn't support it, we need to remove it.

**Would you like to work on the issue?**
yes

amitiwary999:
I want to work on this issue.

chiragw15:
@amitiwary999 please complete your old issues first and then start working on it.

amitiwary999:
@mayank408 I was working on this issue. I already mention this.

mayank408:
as @chiragw15 mentioned I thought you were working on old issues. Also, it was unassigned so I assigned it to myself as I don;t have any other issue to work. Please assign it to yourself first if you want to work on it. Thanks

amitiwary999:
But you didn't ask me. I already started working on this issue. Even I mention it in chat that there is some issue with important message. I asked 2 days before for that. So I already planned for it.

mayank408:
Why would I ask you it was unassigned so I assigned it to me, you could also do so. How would the co-dev know that you "already planned it". Pls work cooperatively.

betterclever:
@amitiwary999 As @chiragw15 said, you had some pending issues. But that aside, your PR seems to not have removed all traces of ImportantMessages. Only 7 changed files in comparison to 19 in PR by @mayank408 . If you started working on this early, you should take care that issue gets resolved fully.

amitiwary999:
@mayank408 Currently I have no right for this that's why I can't assign issues to my self. I only comment that I want to work on this issue. I already mention that.

mayank408:
@amitiwary999 But even after your comment @chiragw15  mentioned that you have to do old issues first and after which there was no response from your side. Well lets be more clear in future as we are all working together. Thanks

amitiwary999:
No problem. Currently I have no right so if I mention that I want to work on this issue atleast ask me once. I can't assign issue to myself.

mayank408:
@amitiwary999 Ya sure will ask you and assign you at the same time if you wants to work on an issue. Cheers 👍

PR #721, opened 2017-06-16
Fix : Removed important messages activity

amitiwary999:
Fixes issue #717


mayank408:
@amitiwary999 I don't understand what are you trying to achieve there is already a pull request regarding this issue. Please work with the community and your co-developers. Firstly the issue was not assigned to you, So you cant make a pull request and above all when there is already a PR does this make any sense.

Issue #723, opened 2017-06-17
Issue with date of retrieved old messages


amitiwary999:
**Actual Behaviour**

Currently date of all retrieved message is same which is date of last message.

**Expected Behaviour

Retrieved message should be shown according to it's date.

**Steps to reproduce it**

Logout and login. See the date of retrieved message. It is same for all. But in response date is different.

**LogCat for the issue**

**Screenshots of the issue**
.

**Would you like to work on the issue?**

Yes.

chiragw15:
@amitiwary999 #712 is same. The reason why there is date after every message because date of all retrieved messages are same. So, I just changed that and it is solved. Sending a PR in some time. Cheers.

chiragw15:
Okay sorry, Didn't see you already sent a PR. Cool.

mayank408:
same as #712. Hence closing

amitiwary999:
@mayank408 It is not same as #712 . #712 is for current chat messages and this issue is related to retrieved old messages after user login again. Please open it.

mayank408:
@amitiwary999 @chiragw15 Pls be more clear. As @chiragw15 said #712 is same as it. I am opening it.

chiragw15:
@amitiwary999 The reason why issue #712 was occurring was that the retrieved message had same date which is this issue. So, the reason of issue #712 is this issue. Thats why I said it is same as #712.

mayank408:
Fixing all date and time related bugs PR #728 is merged. Hence closing.

Issue #712, opened 2017-06-16
Bug: Date is displayed after each message


chiragw15:
**Actual Behaviour**

Date is displayed after every message.

**Expected Behaviour**

No need of displaying date after each message. 

**Screenshots of the issue**

![screenshot_20170616-164012](https://user-images.githubusercontent.com/16978820/27224473-f716f130-52b2-11e7-8a1e-f5d344d4c137.png)

**Would you like to work on the issue?**
yes

mayank408:
Merged hence closing.

PR #724, opened 2017-06-17
Fix : issue with date of retrieved old messages

amitiwary999:
Fixes issue #723 

Screenshots for the change: 
Before change
screenshot_2017-06-17-16-26-53

After Change
screenshot_2017-06-17-17-08-18

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -356,10 +356,19 @@ public class MainActivity extends AppCompatActivity {
                         if(allMessages.size() == 0) {
                             showToast("No messages found");
                         } else {
-                            String date = response.body().getCognitionsList().get(0).getQueryDate();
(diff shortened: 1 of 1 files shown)


amitiwary999:
@chiragw15 @mayank408  please review

chiragw15:
Okay mistakes :
1. There are 2 date formats in app. Keep one. If you compare two same dates of different format using .equals() , they will be different.
2. No consideration of time zone. The date from server is an ISO 8601 date format with zulu time zone which is UTC 0. So basically you are showing dates.
3. Timestamp of retrieved message is timestamp at which it is retrieved and not at which it was queried.
4. Parsing of date in getdate() method is shitty.  
5. Answer Date should be displayed instead of query date.

chiragw15:
Leave these points now. I have already implemented them while solving #712 . Don't implement them again and waste your and my time. Fix some other issue.

amitiwary999:
It is not about current chat. This pr is related to retrieved old messages.

chiragw15:
Already fixed in #728

Issue #725, opened 2017-06-17
Using String builder instead of string


mayank408:
**Actual Behaviour**

String are being used and concatenated using "+" which is not the correct way.

**Expected Behaviour**

Instead of strings , stringBuilder must be used. 

**Screenshots of the issue**

String shareText = "";
                        for (Integer i : getSelectedItems()) {
                            ChatMessage message = getData().get(i);
                            Log.d(TAG, message.toString());
                            shareText += "[" + message.getTimeStamp() + "]";
                            shareText += " ";
                            shareText += message.isMine() ? "Me: " : "Susi: ";
                            shareText += message.getContent();
                            shareText += "\n";
                        }
                        shareText = shareText.substring(0, shareText.length() - 1);
                        shareMessage(shareText);
**Would you like to work on the issue?**

Let us know if this issue should be assigned to you or tell us who you think could help to solve this issue.
Yes

PR #726, opened 2017-06-17
Using StringBuilder instead of string

mayank408:
Fixes issue #725 

StringBuilder is added instead of String during concatenation.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -665,12 +665,13 @@ public class ChatFeedRecyclerAdapter extends SelectableAdapter implements Messag
 
             if (model != null) {
                 List<String> urlList = extractLinks(model.getContent());
-                String url = urlList.get(0);
-                if (!(url.startsWith("http://") || url.startsWith("https://"))) {
-                    url = "http://" + url;
+                StringBuilder url = new StringBuilder(urlList.get(0));
+                StringBuilder http = new StringBuilder("http://");
+                if (!(url.toString().startsWith(http.toString()) || url.toString().startsWith(http.toString()))) {
+                    url = http.append(url.toString());
                 }
                 TextCrawler textCrawler = new TextCrawler();
-                textCrawler.makePreview(linkPreviewCallback, url);
+                textCrawler.makePreview(linkPreviewCallback, url.toString());
             }
         } else {
             linkPreviewViewHolder.titleTextView.setText(model.getWebLinkData().getHeadline());
@@ -1023,8 +1024,7 @@ public class ChatFeedRecyclerAdapter extends SelectableAdapter implements Messag
                                 copyText.append(message.isMine() ? "Me: " : "Susi: ").append(message.getContent()).append("\n");
                             }
                         }
-                        String copyText2 = copyText.substring(0, copyText.length() - 1);
-                        setClipboard(copyText2);
+                        setClipboard(copyText.toString());
                     }
 
                     if (nSelected == 1){
@@ -1057,8 +1057,7 @@ public class ChatFeedRecyclerAdapter extends SelectableAdapter implements Messag
                                 shareText.append(message.isMine() ? "Me: " : "Susi: ").append(message.getContent()).append("\n");
                             }
                         }
-                        String shareText2 = shareText.substring(0, shareText.length() - 1);
-                        shareMessage(shareText2);
+                        shareMessage(shareText.toString());
                     }
 
                     actionMode.finish();
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -31,11 +31,6 @@ import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.Toast;
 
-import com.bumptech.glide.Glide;
-import com.bumptech.glide.RequestManager;
-import com.bumptech.glide.load.resource.drawable.GlideDrawable;
-import com.bumptech.glide.request.RequestListener;
-import com.bumptech.glide.request.target.Target;
 import com.github.mikephil.charting.data.Entry;
 import com.github.mikephil.charting.data.PieData;
 import com.github.mikephil.charting.data.PieDataSet;
@@ -44,6 +39,7 @@ import com.github.mikephil.charting.utils.ColorTemplate;
 import com.leocardz.link.preview.library.LinkPreviewCallback;
 import com.leocardz.link.preview.library.SourceContent;
 import com.leocardz.link.preview.library.TextCrawler;
+import com.squareup.picasso.Picasso;
 
 import org.fossasia.susi.ai.R;
 import org.fossasia.susi.ai.adapters.viewholders.ChatViewHolder;
@@ -104,7 +100,6 @@ public class ChatFeedRecyclerAdapter extends SelectableAdapter implements Messag
     private static final int SEARCH_RESULT = 10;
     private static final int WEB_SEARCH = 11;
     private static final int DATE_VIEW = 12;
-    private final RequestManager glide;
     public int highlightMessagePosition = -1;
     public String query = "";
     private String webquery ;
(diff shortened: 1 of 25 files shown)


mayank408:
@chiragw15 @betterclever @amitiwary999 @chashmeetsingh Pls review.

chiragw15:
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java#L284  
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/activities/SettingsActivity.java#L54
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L671
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L1006
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L1060
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.java#L10
https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.java#L12
@mayank408 Here are some references. Change them here too. I may have missed some, so check once again.

mayank408:
@chiragw15 There is no need to change every String, like the one which are inline. Pls review @iamareebjamal

mayank408:
@chiragw15 It does make sense to change every string to stringBuilder.

chiragw15:
Okay, I am not sure about this. @iamareebjamal can tell better.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1072,20 +1072,20 @@ public void onClick(DialogInterface dialog, int which) {
...
                             ChatMessage message = getData().get(i);
                             if (message.getActionType()==null || message.getActionType().equals(Constant.ANSWER)) {
                                 Log.d(TAG, message.toString());
-                                shareText += "[" + message.getTimeStamp() + "]";
-                                shareText += " ";
-                                shareText += message.isMine() ? "Me: " : "Susi: ";
-                                shareText += message.getContent();
-                                shareText += "\n";
+                                shareText.append("[").append(message.getTimeStamp()).append("]");
+                                shareText.append(" ");
```
iamareebjamal: append is chainable. Chain it like you have done in previous line

iamareebjamal:
@chiragw15 There is no need to do inline strings. In fact, string builder performs poorly than string concatenation for small additions like 2-3 strings

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
mayank408: Changes done @iamareebjamal Please review and approve.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1051,20 +1048,17 @@ public void onClick(DialogInterface dialog, int which) {
...
-                                shareText += message.isMine() ? "Me: " : "Susi: ";
-                                shareText += message.getContent();
-                                shareText += "\n";
+                                shareText.append("[").append(message.getTimeStamp()).append("]").append(" ");
+                                shareText.append(message.isMine() ? "Me: " : "Susi: ").append(message.getContent()).append("\n");
                             }
                         }
-                        shareText = shareText.substring(0, shareText.length() - 1);
-                        shareMessage(shareText);
+                        String shareText2 = shareText.substring(0, shareText.length() - 1);
```
betterclever: Why are you extracting string in this way? You could use shareText.toString()

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1014,20 +1014,17 @@ public void onClick(DialogInterface dialog, int which) {
...
-                                copyText += message.isMine() ? "Me: " : "Susi: ";
-                                copyText += message.getContent();
-                                copyText += "\n";
+                                copyText.append("[").append(message.getTimeStamp()).append("]").append(" ");
+                                copyText.append(message.isMine() ? "Me: " : "Susi: ").append(message.getContent()).append("\n");
                             }
                         }
-                        copyText = copyText.substring(0, copyText.length() - 1);
-                        setClipboard(copyText);
+                        String copyText2 = copyText.substring(0, copyText.length() - 1);
```
betterclever: Here too.

iamareebjamal:
@mayank408 There are other places you can use StringBuilder too. Please search for the places where there are more than 3 concatenations. Thanks :smile:

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1014,20 +1015,16 @@ public void onClick(DialogInterface dialog, int which) {
                         }
                         setClipboard(copyText);
                     } else {
-                        String copyText = "";
+                        StringBuilder copyText = new StringBuilder("");
```
iamareebjamal: No need to add a double quote as an argument in StringBuilder

app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.java
```diff
@@ -6,10 +6,10 @@
 import org.fossasia.susi.ai.model.MapData;
 
 public class MapHelper {
-    private static final String BASE_URL = "http://staticmap.openstreetmap.de/staticmap.php?";
-    private static final String URL_SCHEME = BASE_URL + "center=%f,%f&zoom=%f&size=%sx%s";
-    private static final String MAP_URL = "https://www.openstreetmap.org/#map=";
-    private static final String MAP_URL_SCHEME = MAP_URL + "%f/%f/%f";
+    private static final StringBuilder BASE_URL = new StringBuilder("http://staticmap.openstreetmap.de/staticmap.php?");
+    private static final StringBuilder URL_SCHEME = BASE_URL.append("center=%f,%f&zoom=%f&size=%sx%s");
```
iamareebjamal: These were single concats. Please only use StringBuilder when more than 3 concats are there

betterclever:
@mayank408 There are many places in current diff, in which StringBuilder is used even when there are just concatenation 1-2 times. Is it necessary in all those places? It might be improving performance , if it really is, but decreases code legibility to a great extent there.
So, as @iamareebjamal pointed out, we must do it in places with more than 3 concatenations and not all the places perhaps. 
@iamareebjamal Correct me if I am wrong.

app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.java
mayank408: Sorry but I could not found where there is 3 concats. Thanks

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
mayank408: Okay

iamareebjamal:
@betterclever Correct. Somehow my review did not get posted on github where I said the same. Besides concatenating 2 strings is better than StringBuilder performance wise too. Only use StringBuilder when concatenating more than 3 strings. @chiragw15 has already given many sources where StringBuilder is needed

mayank408:
@iamareebjamal I checked out the resources given by @chiragw15 . The thing we cant use string builder in URI's and in other I have used it.

app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.java
iamareebjamal: Then please fall back to String concat in these places. Thanks

mayank408:
@chiragw15 @iamareebjamal I could not find any use of StringBuilder other than in copy and share method. There is no need to forcefully implement something. So I think we should restrict ourself to that only . Thanks.

betterclever:
@iamareebjamal I see there are some more places, but I am just curious that given that StringBuilder optimizations are introduced in Java 6 compiler which we are targeting, based on our minimum API level like this answers point out.
https://stackoverflow.com/questions/11942368/why-use-stringbuilder-explicitly-if-the-compiler-converts-string-concatenation-t
Do we need it in places without a loop?

iamareebjamal:
Android implementations are different from Oracle JDK implementation. And it is good practice because nevertheless, it stops you from recreating immutable string objects at each concatenation

mayank408:
@iamareebjamal All changes done.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
iamareebjamal: This remains to change. Please remove the constructor parameter

mayank408:
@iamareebjamal done

the-dagger:
@mayank408 Please squash your commits.

mayank408:
@chiragw15 @amitiwary999  Pls approve this.

PR #727, opened 2017-06-17
Unnecessary function removed.

mayank408:
Fixes issue #717 (some methods were left to be removed)

Some function related to important messages were left that are deleted in this PR.


mayank408:
@chiragw15 @betterclever @amitiwary999 @chashmeetsingh Pls review.

Issue #703, opened 2017-06-15
LongclickListener not working.


mayank408:
**Actual Behaviour**

LongClickListener is not working for search result.

**Expected Behaviour**

LongClickListener should work for search result.

**Would you like to work on the issue?**
Yes

chiragw15:
Yes, I deliberately left this earlier. This needs to be thought through. You have created the issue now, so let's discuss this. We need to modify click listeners for websearch, maps and RSS. When we long click any message, there are three possible options : copy that message, share that message and delete that message. So, if the action type is map or websearch , we don't need first 2 options. So, we can simply remove them when action type other than answer and anchor is selected and just keep only delete option. This is inspired from Google assistant. @mayank408 @amitiwary999 views?

mayank408:
@chiragw15 Yes , I totally agree with you. Lets do this way only 👍

amitiwary999:
@chiragw15 :+1:

mayank408:
@chiragw15 I am working on it.

PR #733, opened 2017-06-19
Issue of Longclick listener working on websearch is resolved.

mayank408:
Fixes issue #703 
LongClick listener added for Websearch answer type. 

Screenshots for the change: 

screenshot_1497881113


betterclever:
@mayank408 I am not able to understand that why did you move SearchResultsAdapter to the ChatFeedRecyclerAdapter.java Did implementation there inhibit you from implementing something?

mayank408:
@betterclever I had a long discussion with @chiragw15 regarding this issue, this is best I could do to solve this one. I tried everything but this only worked.
Thanks

betterclever:
No I don't have any problem with implementation but I think same could be achieved if we keep SearchResultsAdapter in a different file. If not, please tell what was not possible.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1093,4 +1083,131 @@ private void shareMessage(String message) {
             currContext.startActivity(sendIntent);
         }
     }
+
+    private class SearchResultsAdapter extends RecyclerView.Adapter<SearchResultHolder> {
```
the-dagger: Why is this an inner class?
This should be kept in a separate class, any special reason as to why you made this an inner class?

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -1093,4 +1083,131 @@ private void shareMessage(String message) {
...
+        private LayoutInflater inflater;
+        private Context context;
+        private List<Datum> datumList;
+        private Realm realm;
+
+
+        private SearchResultsAdapter(Context context, List<Datum> datumList) {
+            this.context = context;
```
the-dagger: Also, no need to pass the context over here, you can do 
 `context = parent.getContext();`
 inside your `onCreateViewHolder()` method.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
mayank408: Actually I could not get the reference of Action mode callback in other class. I am looking into it. If you can suggest a better implementation than this by which I could control the menu action from other class it would be very helpful.
Thanks

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
mayank408: Okay.

chiragw15:
Not needed now. Closing it.

Issue #736, opened 2017-06-19
Pass "websearch" string as a parameter in updateDatabase method


chiragw15:
**Actual Behaviour**

Right now websearch string is used to store query to be searched. And it is global. So, when retrieving messages the previous all websearch strings are updated by the new websearch string while updating database.

**Expected Behaviour**

Pass websearch as a parameter in updateDatabase method.

PR #739, opened 2017-06-19
fix : Adding websearch string as a parameter of the updateDatabase method

chiragw15:
Fixes issue #736 
Changes: Passed websearch query string as a parameter to updateDatabase method because it was changing before database update.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
@@ -150,7 +150,7 @@ public class MainActivity extends AppCompatActivity {
     private int offset = 1;
     private ChatFeedRecyclerAdapter recyclerAdapter;
     private Realm realm;
-    public static String webSearch;
+    private String webSearch;
     private TextToSpeech textToSpeech;
     private BroadcastReceiver networkStateReceiver;
     private ClientBuilder clientBuilder;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -1078,7 +1078,7 @@ public void onFailure(Call<SusiResponse> call, Throwable t) {
 
     private void updateDatabase(final long prevId, final String message, final boolean isDate, final String date, final String timeStamp , final boolean mine,
```
the-dagger: @chiragw15 don't you think it's a good idea to add javadocs for the methods we are using to communicate with the server?

chiragw15: @the-dagger nice idea. Can we configure circle ci script to build javadocs after every PR and push it to `gh-pages` branch?

the-dagger: No, I meant that you should write the documentation for methods in the Java class itself. 
Something like this, 

```
/**
 * Validates a chess move.
 *
 * Use {@link #doMove(int theFromFile, int theFromRank, int theToFile, int theToRank)} to move a piece.
 *
 * @param theFromFile file from which a piece is being moved
 * @param theFromRank rank from which a piece is being moved
 * @param theToFile   file to which a piece is being moved
 * @param theToRank   rank to which a piece is being moved
 * @return            true if the move is valid, otherwise false
 */
boolean isValidMove(int theFromFile, int theFromRank, int theToFile, int theToRank) {
    // ...body
}
```

chiragw15: No, I get that. How else would javadoc be created if we don't write javadoc comments. That will be done. I am asking for hosting javadoc on github pages?

chiragw15: Like this https://ftctechnh.github.io/ftc_app/doc/javadoc/index.html ?

the-dagger: Hosting is not really needed though, but if you like then go ahead.

chiragw15:
ping @mayank408 @amitiwary999

Issue #743, opened 2017-06-20
Search in not working properly 


amitiwary999:
**Actual Behaviour**

For some text like this, here, is search is not working properly

**Expected Behaviour**

Search should work properly.

**Steps to reproduce it**

In the search bar search for this or here or is. You will find that at some places these words are not present still on clicking on up and down button app try to show that word is present. Like if your last query is websearch type then search for 'here' and you find that no text is highlighted.

**LogCat for the issue**

**Screenshots of the issue**

**Would you like to work on the issue?**

Yes.

PR #746, opened 2017-06-20
Fix : search action  is not working properly

amitiwary999:
Fixes issue #743 

Changes: initially answer text (I found this,here is the websearch result etc) was added with websearch,rss, map action type also. So on searching text like 'here', 'this' etc. text found in websearch, rss,map etc but since text is not shown in textview so app not show highlighted text. I changed the code to add answer in anchor and action type only.

Screenshots for the change:


chiragw15:
@amitiwary999 Just adding `answer = null` [here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java#L279) should solve the problem.

amitiwary999:
@chiragw15 @mayank408 review.

Issue #751, opened 2017-06-21
Issue with chatmessage edittext


amitiwary999:
**Actual Behaviour**

Currently after speech input user have to click on edittext if user want to type anything

**Expected Behaviour**

It should not happen

**Steps to reproduce it**

Click on edittext, soft keyboard will open, then click on mic button and give some input. You will find that even soft keyboard is already open but to give text input you have to click on edittext again .

**LogCat for the issue**

**Screenshots of the issue**

**Would you like to work on the issue?**

Yes

PR #752, opened 2017-06-21
Fix : issue with chatmessage edittext

amitiwary999:
Fixes issue #751 

Changes: request focus when voice input hide.


chiragw15:
@amitiwary999 Also, when user writes something in edit text and then press home, the send icon changes to mic icon. Check it once and solved that too in this PR only.

amitiwary999:
@chiragw15 @mayank408 review.

PR #753, opened 2017-06-21
Fix : LongClickListener on Websearch.

mayank408:
Fixes issue #703 
LongClickListener is added.

Screenshots for the change: 

screenshot_1498066288

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/SearchResultsAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/SearchResultsAdapter.java
@@ -1,14 +1,21 @@
 package org.fossasia.susi.ai.adapters.recycleradapters;
 
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
+import android.graphics.Color;
 import android.net.Uri;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.RecyclerView;
 import android.text.Html;
 import android.text.TextUtils;
+import android.util.Log;
+import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.Button;
+import android.widget.Toast;
 
 import com.leocardz.link.preview.library.LinkPreviewCallback;
 import com.leocardz.link.preview.library.SourceContent;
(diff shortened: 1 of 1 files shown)


mayank408:
@the-dagger Is this all right now.

mariobehling:
@mayank408 Please actively ask your co-developers to review using the "@" feature.

I am unsure about the "Delete" feature, cause any feature we have needs to be implemented in the server. It is quite tricky to delete any message, cause this would mean we tamper with the logs. We need to make a number of changes in the server to actually have a delete function.

mayank408:
@mariobehling We are saving the messages in local database of mobile phone and with delete method, we are actually deleting the messages from the local phone database. 

As you are saying that delete function must be implemented it server first, so there can be two things:-

1. We should remove the delete feature from the client.
2. We should open an issue in the Susi server regarding deleting the messages from the server.

@chiragw15 @betterclever @amitiwary999 Pls cmnt.

mariobehling:
> We are saving the messages in local database of mobile phone

We need this to work just like messengers. If I log in to Whatsapp, Facebook web client, Android or where-ever it always shows the same history. Please change this in the way that it gets the history from the server to avoid inconsistencies.

> We should remove the delete feature from the client.

I guess we have to for now.

> We should open an issue in the Susi server regarding deleting the messages from the server.

There was an issue already, but this is currently out of scope as we are not storing data in a DB but in logs for now. For now we should align all clients and make them store things like settings etc. on the server and ensure they are all using the same settings etc.

chiragw15:
@mariobehling 
>Please change this in the way that it gets the history from the server to avoid inconsistencies.

We can not do this now as server only provides 10 messages in chat history. There is an issue regarding this on server https://github.com/fossasia/susi_server/issues/261 to provide complete chat history. Will need to implement that first. After that we'll open issue in all clients to align with it.

chiragw15:
I'll start working on the issue on the server. Meanwhile  @mayank408 can create an issue to remove delete option from the app and work on it.

mariobehling:
@chiragw15 
Ok. So, what happens with this PR?

chiragw15:
If we are removing delete option from android, then no need for it I guess. Can be closed.

amitiwary999:
Yes close this PR if we are removing delete option.

amitiwary999:
@mayank408 I opened an issue [here](https://github.com/fossasia/susi_android/issues/755).

chiragw15:
Not needed now. Closing it.

Issue #755, opened 2017-06-21
Remove delete option from android app


amitiwary999:
**Actual Behaviour**

Currently chat messages are not saved in server db . So deleting chat messages not involve server.

**Expected Behaviour**

Since server doesn't support it we need to remove it.

**Would you like to work on the issue?**

Yes.

amitiwary999:
I will work on this issue.

PR #758, opened 2017-06-22
Fix : Removed delete option.

mayank408:
Fixes issue #755

Delete option is removed from the app as it is not implemented in server.


mayank408:
@amitiwary999 @chiragw15 Pls revew.

Issue #701, opened 2017-06-15
Text to speech is not working properly


amitiwary999:
**Actual Behaviour**

tts is not working properly .

**Expected Behaviour**

tts should work properly.

**Would you like to work on the issue?**

yes. 

@chiragw15 @mayank408 can you test it. I tried many times but it is not working properly.

chiragw15:
@amitiwary999 Can you provide a little detail in the issue like
1. Which device you are using for testing, 
2. Logs if any,
3. When do you face this issue,
4. Is it not working at all or for some particular queries it doesn't work. 
5. Steps to reproduce it like which particular query should we enter.

Simply writing "tts is not working" and "tts should work" won't help.

amitiwary999:
I tried it on Lenovo A6000 plus lollipop and Asus Zenfone marshmallow.
It is not working properly for any query. Sometimes it is very late and
sometimes it is not working.

chiragw15:
Yup seems like the problem arises after map or rss queries. Looking into it. Will let you know if I find the cause.

amitiwary999:
@chiragw15 There is a handler in computeOtherMessage  and from this handler we are calling voiceReply and there is also a handler.When I removed this handler TTS is working perfectly.

amitiwary999:
I tried app on several devices with several account. Sometimes susi
replying late for rss, websearch and map action type, app is doing some
extra task for parsing and showing data in case of map, websearch that's
why tts is delayed. Have you find some solution?

mayank408:
@amitiwary999 Could you tell me how to reproduce the error.

amitiwary999:
Query some websearch , map or rss action type reply.

PR #759, opened 2017-06-22
Fix : tts is not working properly

amitiwary999:
Fixes issue #701


amitiwary999:
@chiragw15 can you try this. I tried few solutions. This solution seems to work properly.

amitiwary999:
@chiragw15 @mayank408  review

Issue #760, opened 2017-06-22
UI issue with web search


mayank408:
**Actual Behaviour**

No right margin is assigned to web search
**Expected Behaviour**

Right margin must be assigned to web search.

**Screenshots of the issue**

![screenshot_1498124522](https://user-images.githubusercontent.com/22056321/27427689-7e18ac6a-575d-11e7-9c1b-46a969f945cf.png)

**Would you like to work on the issue?**

Yes

PR #761, opened 2017-06-22
Fix : Added margin to the websearch.

mayank408:
Fixes issue #760 

Right margin is added to the websearch.
Screenshots for the change:

screenshot_1498125142


mayank408:
@amitiwary999 Yes it was intentional like the left and right margin will have same width while the gap between the cards will be double of it. In this way it is uniform.

amitiwary999:
Margin should be uniform everywhere like in allo app.

amitiwary999:
@mayank408 I think this issue is related to both rss and websearch action type.

mayank408:
@amitiwary999 As suggested by you I have made margin equal.

screenshot_1498141224

mayank408:
@chiragw15 @amitiwary999 @the-dagger Pls review.

Issue #742, opened 2017-06-20
App crashes after logout 


amitiwary999:
**Actual Behaviour**

App is crashing after logout 

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

logout from app. Login and logout 

**LogCat for the issue**

Process: org.fossasia.susi.ai, PID: 29039
                                                                      java.lang.IllegalStateException: Realm instance has been closed or this object or its parent has been deleted.
                                                                          at io.realm.RealmList.checkValidView(RealmList.java:754)
                                                                          at io.realm.RealmList.size(RealmList.java:548)
                                                                          at org.fossasia.susi.ai.adapters.recycleradapters.WebSearchAdapter.getItemCount(WebSearchAdapter.java:111)
                                                                          at android.support.v7.widget.RecyclerView.dispatchLayoutStep1(RecyclerView.java:3416)
                                                                          at android.support.v7.widget.RecyclerView.onMeasure(RecyclerView.java:2963)
                                                                          at android.view.View.measure(View.java:17438)
                                                                          at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:5465)
                                                                          at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1436)
                                                                          at android.widget.LinearLayout.measureVertical(LinearLayout.java:722)
                                                                          at android.widget.LinearLayout.onMeasure(LinearLayout.java:613)
                                                                          at android.view.View.measure(View.java:17438)
                                                                          at android.support.v7.widget.RecyclerView$LayoutManager.measureChildWithMargins(RecyclerView.java:8456)
                                                                          at android.support.v7.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1551)
                                                                          at android.support.v7.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1488)
                                                                          at android.support.v7.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:570)
                                                                          at android.support.v7.widget.RecyclerView.dispatchLayoutStep2(RecyclerView.java:3506)
                                                                          at android.support.v7.widget.RecyclerView.dispatchLayout(RecyclerView.java:3254)
                                                                          at android.support.v7.widget.RecyclerView.consumePendingUpdateOperations(RecyclerView.java:1593)
                                                                          at android.support.v7.widget.RecyclerView$ViewFlinger.run(RecyclerView.java:4625)
[...]

PR #777, opened 2017-06-27
Fix : app crashes after logout

amitiwary999:
Fixes issue #742 

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -53,6 +53,7 @@ import org.fossasia.susi.ai.helper.AndroidHelper;
 import org.fossasia.susi.ai.helper.Constant;
 import org.fossasia.susi.ai.helper.ConstraintsHelper;
 import org.fossasia.susi.ai.helper.MapHelper;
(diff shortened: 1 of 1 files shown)


amitiwary999:
@mayank408 I explained reason [here](https://github.com/fossasia/susi_android/issues/742). Here I am checking if user is logout(after logout token expire and realm closed) then not perform linkpreview work(background work).

chiragw15:
@amitiwary999 You said in #742 that
>Reason for crash is after logout realm closed but linkpreview still running in background and it try to change real object but realm is already closed.

If this is the case, shouldn't the app crash too when app is closed because realm instance is closed in onDestroy mehtod. ([here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java#L1605))

chiragw15:
@the-dagger I have a doubt. Is linkPreviewCallback ([here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java#L620)) executed after main activity is closed. Or in general are call backs executed if the app is closed or the activity is destroyed?

amitiwary999:
@chiragw15 in onDestroy we are not deleting realm. But in case of logout we are deleting realm objects. I think that's why app is not crashing. In logcat it is shown that ' Object is no longer valid to operate on.Was it deleted by another thread?'

Issue #779, opened 2017-06-29
App is crashing.


mayank408:
**Actual Behaviour**

App is crashing see below for more details.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

Try to change wallpaper or mic settings.

**LogCat for the issue**

Provide logs for the crash here

![screen shot 2017-06-30 at 1 21 11 am](https://user-images.githubusercontent.com/22056321/27707780-2f5c4cb2-5d34-11e7-8ef7-66fbf551f21e.png)

**Would you like to work on the issue?**

Yes

PR #780, opened 2017-06-29
Fix : Crash due to onRequest Permission method.

mayank408:
Fixes issue #779 

Changes: Changed the check inside the if statement.


mayank408:
@chirag15 @amitiwary999 Pls review

Issue #766, opened 2017-06-22
Store User settings on server


chiragw15:
Store the user settings on the server and push and pull from the server accordingly to maintain the app state.

Refer fossasia/susi_server#306 for endpoints to push data to server

Listing Settings :
http://api.susi.ai/aaa/listUserSettings.json?access_token=ACCESS_TOKEN

Adding Settings :
http://api.susi.ai/aaa/changeUserSettings.json?key=theme&value=THEME&access_token=ACCESS_TOKEN

mayank408:
http://api.susi.ai/aaa/changeUserSettings.json?key=theme&value=THEME&access_token=ACCESS_TOKEN

The link is not working.

mayank408:
Which theme are we gonna store in the server?

chiragw15:
Where is the mention of theme in the issue? I have specifically mentioned `user settings`. If you are talking about theme in the link, that is just an example. Another example can be http://api.susi.ai/aaa/changeUserSettings.json?key=voice_input&value=true&access_token=ACCESS_TOKEN

mayank408:
yes , I was talking about the use cases.

chiragw15:
All the settings like voice_input , voice_output, voice_output_on_voice_input will be stored on server. Apart from that, in future if we also have different themes : light and dark like http://chat.susi.ai/, those will also be stored on server.

PR #781, opened 2017-06-29
feat : store user settings on server

amitiwary999:
Fixes issue #766

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/activities/SettingsActivity.java
+++ b/app/src/main/java/org/fossasia/susi/ai/activities/SettingsActivity.java
@@ -29,8 +29,13 @@ import org.fossasia.susi.ai.helper.CredentialHelper;
 import org.fossasia.susi.ai.helper.MediaUtil;
 import org.fossasia.susi.ai.helper.PrefManager;
 import org.fossasia.susi.ai.login.LoginActivity;
+import org.fossasia.susi.ai.rest.ClientBuilder;
+import org.fossasia.susi.ai.rest.responses.susi.ChangeSettingResponse;
 
 import io.realm.Realm;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
 
 import static org.fossasia.susi.ai.helper.Constant.DARK;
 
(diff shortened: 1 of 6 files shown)


mayank408:
@amitiwary999 I think its too early to implement this. Its needs to be discussed what settings and preferences needs to be stored in the server. And these must be similar in all the clients. Also you have assigned the issue neither commented that you are working on it, so that we could first discuss how we going to approach this. 
@chiragw15 What do you think..?

chiragw15:
@mayank408 I had a talk with @uday96 . He is also working on the same thing on web client. So, it is not a problem implementing it right now.

So, the settings for now can be : 
1. "enter_send"  :    value can be true/false    :    Will be used for inputing multi line queries. True means pressing enter will send message. False means pressing enter adds a new line.
2. "speech_output"   :    value can be true/false   :    Will be used for enable speech output when input type is speech.
3. "speech_always"   :    value can be true/false    :    Will be used to enable speech output irrespective input type.
4. "mic_input"        :        value can be true/false    :    Will be used to enable speech input.
5.  "theme"            :        value will be string (dark or light).     Will be used to change theme of user. 

Usage will be to get them from server during login. And saving them to server during log out.
@uday96 @chashmeetsingh views on this?

chashmeetsingh:
Yes 👍🏻 I guess the colour scheme can also be stored on server for the themes, what do you think?

uday96:
These key value pairs seem good to store user settings on server. :+1: 
For theme , in webchat its 'light' or 'dark' as of now. We can use colour schemes too, but  for that we will have to first figure out the components and have an object with the components as keys and colours as values. But 'light' or 'dark' should be fine for now I guess. We can later change it into colour schemes if required (Like if we are giving an option for custom themes too - Mario told that it isn't a priority as of now).

amitiwary999:
Yes currently theme is not implemented in android. once it is implemented I will include it to store on server.

chiragw15:
@amitiwary999 You need to make some changes. First you have to fetch user settings after login before starting main activity. Implement it in a way to fetch user settings at the time of login.
If( login == success) fetch user settings and start main activity. Secondly, only send settings to server when user change it and not when it is changed automatically due to absence of STT. Got my point?

chiragw15:
@uday96 @chashmeetsingh right now, lets only keep theme as dark/light. We can add customizing of theme later on.

chashmeetsingh:
Yup 👍

amitiwary999:
@chiragw15 what do you mean by setting changes automatically? You mean STT setting changes automatically according to device so not send it to server when it changes automatically?

amitiwary999:
@chiragw15 @mayank408 review

chiragw15:
@amitiwary999 Yes. When user change setting himself, then only send it to server. Also, do this in background so user doesn't have to wait. Send mic and theme settings too. Just make the changes and after PR for theme settings gets merged, we can merge this.

amitiwary999:
@chiragw15 @mayank408 review

amitiwary999:
will reopen this pr and start working on it when issue [828](https://github.com/fossasia/susi_android/issues/828) is solved otherwise it will create large conflicting files.

Issue #817, opened 2017-07-08
Implement onDetach Method in Login and Signup Activities


mayank408:
**Actual Behaviour**

Implement onDetach Method is not implemented in Login and Signup Activities

**Expected Behaviour**

Implement onDetach Method in Login and Signup Activities

**Would you like to work on the issue?**

Yes

PR #818, opened 2017-07-08
Implemented onDetach Method in Login and Signup Activities

mayank408:
Fixes issue #817

Changes: Implemented onDetach Method in Login and Signup Activities to avoid memory leaks.


mayank408:
@betterclever @chiragw15 @amitiwary999 Pls review.

PR #821, opened 2017-07-10
fix: Some changes in MVP Design of login activity plus refracting

### LINKED ISSUE #822: Some improvement in MVP design
I asked for feedback from someone on the new MVP design we are following. Although, we are going in a right direction, there were some mistakes that we made like we were certain about not adding business logic in view but we missed the point that we should also not make UI update from model and we called showProgress() method from model which was supposed to be called from presenter. 

Some points I got back as feedback : 
1. Seems like interactor is only used as a proxy to view, this is bad design. You are just using it to call functions in presenter.
2. The presenter's work is major after model's work. All logic should be handled there itself except the model (data work). Data access should be in presenter. But data access logic and handling should be in Model.
3. 
```
fun incorrectEmailError()
fun emptyEmailError()
fun emptyPasswordError()
fun emptyURLError()
fun invalidURLError()
```
These error functions can be reduced to single error function.
4. 
```
fun showProgress()
fun hideProgress()
```
This can be changed to showProgress(boolean) where true or false indicates either to show or false
5. Model will make the request, save in db if required, handle errors
Presenter will only call model's method and receive either success -> Data or error -> Message
In short for getting data presenter -> model
For UI presenter -> View
For any interaction View -> Presenter
6.  There are certain things, your presenter isn't doing much
It's just a proxy for interactor
7. Interactor should not command presenter to show or hide progress bar
In fact, interactor should be oblivious to UI
Your login function is too complex,
It should only provide listener with 2 return types
Success or error
And then presenter will parse error types and show appropriate UI errors
8. Another thing, presenter should be validating data, not model
Model's work is to only request and provide data, presenter is the intermediator
Which processes the data, validates before sending
Receives and processes again, shows any warning, display empty messages or the data

I have opened a PR #821 regarding this which changes the above mentioned points in login.
Also it seemed like Interactor can be a little confusing, so I used old school approach and renamed it to Model. Apart from that, I refracted the code and made some packages : a different package for all models, a contract package for interfaces. @mayank408 @amitiwary999 Please go through that PR and see the changes and accordingly implement in sign up, forgot password and settings.

chiragw15:
Fixes #822 
Changes: 
1. Improved login model and presenter.
2. Refracted the code and made new packages for models and contract
3. Made a new UtilModel for utilities like preferences.


chiragw15:
Update : 
1. Initialized models in constructor rather than onAttach() method.
2. Removed nullable variables. Used `lateinit` property.

chiragw15:
@amitiwary999 @mayank408 Please review this. After that change the forgot password and sign up functionalities accordingly and follow the similar directory structure. There is some common code changes in it like UtilModel, New Directory structure(making separate package for models) which will be required for other features as well.

app/src/main/java/org/fossasia/susi/ai/login/LoginPresenter.kt
```diff
@@ -1,85 +1,132 @@
 package org.fossasia.susi.ai.login
 
 import android.content.Context
+import android.util.Patterns
```
amitiwary999: I found [here](https://medium.com/@cervonefrancesco/model-view-presenter-android-guidelines-94970b430ddf) that presenter doesn't depend on android classes.

amitiwary999: I found [here](https://medium.com/@cervonefrancesco/model-view-presenter-android-guidelines-94970b430ddf) that presenter doesn't depend on android classes.

chiragw15:
@amitiwary999 Thanks for pointing out. Done changes. Please review.

app/src/main/java/org/fossasia/susi/ai/data/contract/IUtilModel.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.fossasia.susi.ai.rest.responses.susi.LoginResponse
+import retrofit2.Response
+
+/**
+ * Interface for Util Model
+ *
+ * Created by chiragw15 on 10/7/17.
+ */
+interface IUtilModel {
```
amitiwary999: I think IUtilModel will be different for login, signup and forgotpassword? I should be named accordingly. Like ILoginUtilModel.

chiragw15: No. It is "utility" model i.e common for every feature. Just like a util class which can be used from anywhere. It can be used from any presenter. It basically is used to work on context based utility functions like getting strings from resources, getting preferences etc.

amitiwary999: okay

Issue #830, opened 2017-07-11
App is crashing in case on rss action type.


amitiwary999:
**Actual Behaviour**

App is crashing in case of rss action type if description provided by server is null.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

query like 'ffgh'

**LogCat for the issue**

java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
                                                                          at org.fossasia.susi.ai.adapters.recycleradapters.SearchResultsAdapter.onBindViewHolder(SearchResultsAdapter.java:63)
                                                                          at org.fossasia.susi.ai.adapters.recycleradapters.SearchResultsAdapter.onBindViewHolder(SearchResultsAdapter.java:30)
                                                                          at android.support.v7.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:6279)
                                                                          at android.support.v7.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:6312)
                                                                          at android.support.v7.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:5258)
                                                                          at android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:5521)
                                                                          at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5363)
                                                                          at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5359)
                                                                          at android.support.v7.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2141)
                                                                          at android.support.v7.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1525)
                                                                          at android.support.v7.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1488)
                                                                          at android.support.v7.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:585)
                                                                          at android.support.v7.widget.RecyclerView.dispatchLayoutStep2(RecyclerView.java:3506)
                                                                          at android.support.v7.widget.RecyclerView.onMeasure(RecyclerView.java:2969)
                                                                          at android.view.View.measure(View.java:17438)
                                                                          at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:5465)
[...]

PR #831, opened 2017-07-11
Fix : app is crashing in case of rss action type

amitiwary999:
Fixes issue #830


amitiwary999:
@chiragw15 @mayank408 please review

Issue #834, opened 2017-07-12
HTML tags are not parsed in susi android


saurabhjn76:
**Actual Behaviour**

Currently, HTML tags are not getting parsed  in susi
![image](https://user-images.githubusercontent.com/9781788/28128757-797f1b70-674e-11e7-9938-514fafbaefea.png)

**Expected Behaviour**
Like web client HTML should be parsed
![image](https://user-images.githubusercontent.com/9781788/28128941-0773ff40-674f-11e7-90e3-1b9591724fe0.png)

**Would you like to work on the issue?**

@chiragw15 @mayank408 @amitiwary999 Please help in fixing it

amitiwary999:
I will fix it.

amitiwary999:
@saurabhjn76 I tried but response is different. Can you tell me how you get that response?

![screenshot_2017-07-12-23-01-37](https://user-images.githubusercontent.com/14003244/28131000-2749bbfa-6756-11e7-88d8-ced93c7a220c.png)

chiragw15:
https://github.com/fossasia/susi_server/pull/367 PR is still open. Pull the code and run on local host

saurabhjn76:
@amitiwary999 It was from my local server, Please refer to https://github.com/fossasia/susi_server/pull/367 for changes. Thanks

amitiwary999:
Okay

PR #835, opened 2017-07-12
Fix : html tags are not parsed correctly in case of answer action type

amitiwary999:
Fixes issue #834


amitiwary999:
will reopen it later after merging of opened prs.

amitiwary999:
screenshot_2017-07-14-00-17-21

PR #838, opened 2017-07-13
Converted Models to Kotlin from Java

mayank408:
Fixes issue  #839

Converted Models to Kotlin from Java


app/src/main/java/org/fossasia/susi/ai/forgotPassword/IForgotPasswordInteractor.kt
```diff
@@ -0,0 +1,23 @@
+package org.fossasia.susi.ai.forgotPassword
+
+import android.content.Context
+
+/**
+ * Created by meeera on 7/7/17.
+ */
+interface IForgotPasswordInteractor {
```
chiragw15: What is this file doing in this PR?

app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
```diff
@@ -53,7 +53,7 @@ class LoginActivity : AppCompatActivity(), ILoginView {
         loginPresenter.onAttach(this)
     }
 
-    override fun onLoginSuccess(message: String) {
+    override fun onLoginSuccess(message: String?) {
```
chiragw15: Any particular reason to this?

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * <h1>POJO class to parse related topics object in retrofit response from websearch client.</h1>
+
+ * Created by mayank on 09-12-2016.
+ */
+
+class RelatedTopics
+
+(@SerializedName("FirstURL")
+ @Expose
```
chiragw15: What does `@Expose` do?

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
```diff
@@ -0,0 +1,29 @@
...
+ var text: String?,
+
+ @SerializedName("Icon")
+ @Expose
+ var icon: WebIcon?) {
+
+    @SerializedName("Result")
+    @Expose
+    var result: String? = null
+}
```
chiragw15: Why are other three class members separated from this? The result variable is in {} whereas rest are in ()

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/WebSearch.kt
```diff
@@ -0,0 +1,20 @@
...
+
+class WebSearch
+
+(@SerializedName("Heading")
+ @Expose
+
+ var heading: String?, @SerializedName("RelatedTopics")
+ @Expose
+
+ var relatedTopics: List<RelatedTopics>?)
```
chiragw15: Please properly refract this. These things should not be pointed out.

app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/LoginResponse.kt
```diff
@@ -0,0 +1,31 @@
...
+    val session: Session? = null
+
+    @SerializedName("valid_seconds")
+    @Expose
+    val validSeconds: Long = 0
+
+    @SerializedName("access_token")
+    @Expose
+    var accessToken: String? = null
+        internal set
```
chiragw15: What does `internal set` do?

app/src/main/java/org/fossasia/susi/ai/forgotPassword/IForgotPasswordInteractor.kt
mayank408: It was causing conflict so I added this

app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
mayank408: Yes because without "?" , I have to append "!!" at the end of the string while calling the function

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
mayank408: It is same as the java annotation we use for retrofit parsing of objects into class

chiragw15: I can see that but what does it do?

mayank408: https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/annotations/Expose.html

Refer to this

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
mayank408: Look at the constructor in java class pls once it have three member classes that's why it is in the constructor of the kotlin class while result is inside the class {} and not in the constructor

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/WebSearch.kt
mayank408: okay

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
chiragw15: I am asking why are other three in the constructor and the fourth one not. Why is there even a constructor. Is it being used somewhere? If not, no need.

app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
chiragw15: It was working earlier without a "!!" at the end. Any change which generates a need to add a "?" or a "!!"

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
mayank408: For your kind information every POJO class have constructor along with getters and setters but what kotlin do is reduce the getters and setter to the constructor only so it acts a getter and setter function. Pls first read the documentation.

app/src/main/java/org/fossasia/susi/ai/forgotPassword/IForgotPasswordInteractor.kt
chiragw15: Please update the fork and resolve conflicts and push it. This file is delete now and should not exist anywhere.

app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/LoginResponse.kt
mayank408: internal set is one of the modifiers like that of public, private and protetected. It can be used to access the variable at same package level. 

Refer to this 
https://kotlinlang.org/docs/reference/visibility-modifiers.html

app/src/main/java/org/fossasia/susi/ai/forgotPassword/IForgotPasswordInteractor.kt
mayank408: When i was working on it , it was not merged. Okay i will remove it.

app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
mayank408: Yes it was causing error after some changes changes so I changed to this approach.

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/RelatedTopics.kt
chiragw15: As far as I know, there is no need of explicitly defining a no argument constructor. The compiler does that for you. 
https://stackoverflow.com/questions/3641114/do-i-really-need-to-define-default-constructor-in-java
According to 'documentation' https://docs.oracle.com/javase/tutorial/java/javaOO/constructors.html 
>You don't have to provide any constructors for your class, but you must be careful when doing this. The compiler automatically provides a no-argument, default constructor for any class without constructors. 

Correct me if I am wrong but all these POJO classes are working without defining a constructor explicitly. https://github.com/fossasia/susi_android/tree/development/app/src/main/java/org/fossasia/susi/ai/rest/responses/susi 

Plus I am just asking, you haven't explicitly used () these brackets anywhere else like 
https://github.com/fossasia/susi_android/pull/838/files#diff-3d6fc652d7495b5553ae8e938dd29431
https://github.com/fossasia/susi_android/pull/838/files#diff-6829e258eaaa5121e5ff77cc43e26b8d
https://github.com/fossasia/susi_android/pull/838/files#diff-ddd769f6bd55d9cc0b3515c5dc4319ce
https://github.com/fossasia/susi_android/pull/838/files#diff-30f6b7e474a3eea82dbd03ddee3fb2b7
https://github.com/fossasia/susi_android/pull/838/files#diff-2ae65d763b750ec12c30376a97ccbddd
https://github.com/fossasia/susi_android/pull/838/files#diff-e9a443de7517d5d06c21bc5b85588fdd
https://github.com/fossasia/susi_android/pull/838/files#diff-d2eb5bc4d9257b98916861793a168a62
https://github.com/fossasia/susi_android/pull/838/files#diff-f34134b4b2b04e01a94ba8dfe11ad340
etc. Why is this class special?

According to me, we don't need to specifically use () brackets here. Then again, correct me if I am wrong.

mayank408: We don't need to define constructor explicitly for java but for kotlin we do

mayank408: () are used only when its a part of constructor otherwise {} are used for members inside the class, I dont understand what causing so much confusion for you in this.

mayank408: The above classes you mentioned does have a construtor so thats why i have used () , all memebers are inside the class simple as that

chiragw15: You are not getting me. I am just can we change 
```
(@SerializedName("FirstURL")
 @Expose
 var url: String?,

 @SerializedName("Text")
 @Expose
 var text: String?,

 @SerializedName("Icon")
 @Expose
 var icon: WebIcon?) {

    @SerializedName("Result")
    @Expose
    var result: String? = null
}
```
To
```
@SerializedName("FirstURL")
 @Expose
 var url: String?

 @SerializedName("Text")
 @Expose
 var text: String?

 @SerializedName("Icon")
 @Expose
 var icon: WebIcon? 

@SerializedName("Result")
@Expose
var result: String? = null
}
```
If not, what is the problem?

chiragw15: >The above classes you mentioned does have a constructor so thats why i havent used () , all members are inside the class simple as that

There is no need of constructor here as well. That's what I told you 3 comments before.

mayank408:
```
public class VideoGame {

    private String name;
    private String publisher;
    private int reviewScore;

    public VideoGame(String name, String publisher, int reviewScore) {
        this.name = name;
        this.publisher = publisher;
        this.reviewScore = reviewScore;
    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public String getPublisher() {
        return publisher;
    }

    public void setPublisher(String publisher) {
        this.publisher = publisher;
    }

    public int getReviewScore() {
        return reviewScore;
    }

    public void setReviewScore(int reviewScore) {
        this.reviewScore = reviewScore;
    }
```

mayank408: is converted to

mayank408: `data class VideoGame(val name: String, val publisher: String, var reviewScore: Int)`

mayank408: in kotlin

mayank408: thus we dont need getters and setters simply a constructor will do the job thats what is happening in this case as well

chiragw15: Mayank, I get that but there is no need of doing that. You can simply change
```
package org.fossasia.susi.ai.rest.responses.others

import com.google.gson.annotations.Expose
import com.google.gson.annotations.SerializedName

/**
 * <h1>POJO class to parse related topics object in retrofit response from websearch client.</h1>
 * Created by mayank on 09-12-2016.
 */

class RelatedTopics

(@SerializedName("FirstURL")
 @Expose
 var url: String?,

 @SerializedName("Text")
 @Expose
 var text: String?,

 @SerializedName("Icon")
 @Expose
 var icon: WebIcon?) {

    @SerializedName("Result")
    @Expose
    var result: String? = null
}
``` 
to
```
package org.fossasia.susi.ai.rest.responses.others

import com.google.gson.annotations.Expose
import com.google.gson.annotations.SerializedName

/**
 * <h1>POJO class to parse related topics object in retrofit response from websearch client.</h1>
 * Created by mayank on 09-12-2016.
 */

class RelatedTopics {

@SerializedName("FirstURL")
 @Expose
 var url: String?

 @SerializedName("Text")
 @Expose
 var text: String?

 @SerializedName("Icon")
 @Expose
 var icon: WebIcon?

@SerializedName("Result")
@Expose
var result: String? = null
}
``` 

What is the problem in doing that. We don't need the constructor.

mayank408: @chiragw15 How would you initialize and set the value of variables in this if needed?

chiragw15: Then, why haven't you done like that in these 
https://github.com/fossasia/susi_android/pull/838/files#diff-3d6fc652d7495b5553ae8e938dd29431
https://github.com/fossasia/susi_android/pull/838/files#diff-6829e258eaaa5121e5ff77cc43e26b8d
https://github.com/fossasia/susi_android/pull/838/files#diff-ddd769f6bd55d9cc0b3515c5dc4319ce
https://github.com/fossasia/susi_android/pull/838/files#diff-30f6b7e474a3eea82dbd03ddee3fb2b7
https://github.com/fossasia/susi_android/pull/838/files#diff-2ae65d763b750ec12c30376a97ccbddd
https://github.com/fossasia/susi_android/pull/838/files#diff-e9a443de7517d5d06c21bc5b85588fdd
https://github.com/fossasia/susi_android/pull/838/files#diff-d2eb5bc4d9257b98916861793a168a62
https://github.com/fossasia/susi_android/pull/838/files#diff-f34134b4b2b04e01a94ba8dfe11ad340

mayank408: Could you pls tell me why would need to intiliaze the above you sent. We neither have constructor for these in java classes nor in kotlin. But apart of it we have constructor for Web in java as well, thus it may be require to intialize it if needed that why it is there in Kotlin.

mayank408: And its not the thing which requires this much discussion I have converted as it was in java

chiragw15: "It was just like that in java" does not mean it was right. If you don't need to initialize the above files I sent, then you don't need to initialize this one either. The values are automatically converted by Retrofit from the JSON object and assigned to variables. There is **no** need to explicitly initialize the variables. Hence we don't need a constructor.

mayank408: We do have constructors for Webicon web search as well, What you think of those :/

chiragw15: Remove from them too. There are 16 POJO classes in response package. Out of which only 4 has a constructor: WebIcon, WebSearch, RelatedTopics and SusiBaseUrls. Out of which, constructor in SusiBaseUrls is actually getting used whereas rest three have useless constructor and coincidently those three POJO classes ( WebIcon, WebSearch, RelatedTopics ) are made by you only. And suppose lets say you may even need to initialize the variables, then why would you write a constructor to initialize 3 out of 4 variables. What has 4th variable done to you? Please, let's end the discussion. Just like other 12 POJO classes, please don't write a constructor in these 3 until and unless it is actually getting used like SusiBaseUrls.

mayank408: okk be happy :)

chiragw15:
@mayank408 How about we use Moshi convertor factory for auto-mapping https://github.com/square/moshi. By this way, we can reduce 
```
class LocationResponse {

    @SerializedName("ip")
    var ip: String? = null

    @SerializedName("hostname")
    var hostname: String? = null

    @SerializedName("city")
    var city: String? = null

    @SerializedName("region")
    var region: String? = null

    @SerializedName("country")
    var country: String? = null

    @SerializedName("loc")
    var loc: String? = null

    @SerializedName("org")
    var org: String? = null

    @SerializedName("postal")
    var postal: String? = null
}
```
to
```
class LocationResponse {
    var ip: String? 
    var hostname: String?
    var city: String? 
    var region: String? 
    var country: String? 
    var loc: String?
    var org: String? 
    var postal: String?
}
```
What say? By this way, we can even fit every susi response in single file, making overall 3 files i.e for susi response, for websearch and for location. See this tutorial for example.

chiragw15:
Okay, I'll make a separate issue for that.

Issue #849, opened 2017-07-15
Converting Helper methods used from Java to Kotlin.


mayank408:
parent issue #787

Converting Helper methods used from Java to Kotlin.

Issue #787, opened 2017-07-03
Shift to Kotlin


chiragw15:
If we plan to go with the MVP design pattern mentioned in #786 , we might need to implement many new classes. So, if we are writing them again, why don't we write them in Kotlin. There is no doubt that Kotlin is better than java with less amount of code for same thing, free from null pointer exception, its interoperability with java, etc. (More information : https://code.tutsplus.com/articles/java-vs-kotlin-should-you-be-using-kotlin-for-android-development--cms-27846 )

It is also now officially supported by google now.

The best part is we don't need to change complete code to kotlin at once. Kotlin and Java works together in same app and user will not even find a difference. So, we can slowly shift to kotlin changing components one at a time and keeping the app always running. 

@mariobehling @the-dagger @mayank408 @amitiwary999 @betterclever views?

betterclever:
Indeed @chiragw15 , kotlin will be a good switch for the app. Code will become much concise. As you mentioned already, best part is that we can shift in phases. I have made an app entirely in Kotlin and I will definitely recommend it.

the-dagger:
Using kotlin sounds good.
Start off by using it in small places and increasing it gradually.
Automatica java to kotlin conversion is available in the canary built, but you can use the kotlin plugin for the stable Studio version.

Issue #786, opened 2017-07-03
Shift to MVP Design Pattern


chiragw15:
The code in the app is already getting pretty messy even though there is not much features implemented. In the future when there will be a lot of action types from server, it will be very difficult to manage code base. This also makes testing of the app extremely difficult and causes various crashes and memory leaks here and there. 

Suggestion -> Using MVP design pattern with dagger 2 injection
 
@iamareebjamal 's blog http://blog.fossasia.org/the-joy-of-testing-with-mvp/ explains very nicely how MVP design pattern works and how to test the app. Also, this https://stackoverflow.com/questions/40766185/advantage-of-mvp-in-android stackoverflow answer lists advantages of using MVP design pattern. 

Since, the codebase is small right now and will get huge with the development of more and more skills on server, it may be a wise decision to shift to MVP architecture.

@mariobehling @the-dagger @mayank408 @amitiwary999 @betterclever views on this?

amitiwary999:
@chiragw15 Yes using MVP architecture is good. I have tried mvp architecture.

mayank408:
@chiragw15 @betterclever I too have been the supporter of MVP design pattern but there few things to keep in mind:-

1. It's quite confusing to understand so for new developers contributing to the code base will find it very difficult to start with.

2. After converting into MVP the codebase will become quite large as compared to the present scenario.

chiragw15:
@mayank408 We can't compromise with the performance of the app just because it is a little difficult to understand. Just a normal 1 day study is enough to get started with MVP. And as far as codebase is concerned we'll follow different packages style. Each feature/activity will have a different package containing everything related to it. If code is organized, it does not matter how large it is.

mayank408:
I think @the-dagger Will be better to answer this question.

iamareebjamal:
You can look at MVVM using new Architecture components as it is a little easier, and solves problems like reactivity and orientation changes easily

mayank408:
@iamareebjamal Yes MVVP seems better than MVP as its is relatively new but I am not sure whether there are enough resources available for it to get started with.

iamareebjamal:
There are several code labs for Architecture Components, several articles on medium and hours of talks from Google I/O on YouTube, and it's easy to convert from any design pattern

the-dagger:
@chiragw15 @mayank408 We should definitely shift to MVP asap, that will make managing the code a lot easier.
I don't really find MVVM that appealing in terms of code readability so let's skip that.

[...]

PR #850, opened 2017-07-15
Changed helper methods to Kotlin.

mayank408:
Fixes issue #849 

I have not converted Constants and PrefManager classes as they are causing conflicts with that of MainActivity we can convert it once MainActivity is converted to Kotlin.


app/src/main/java/org/fossasia/susi/ai/activities/MainActivity.java
```diff
@@ -332,7 +331,7 @@ public static boolean checkSpeechAlwaysPref() {
      * @return the boolean
      */
     public boolean checkMicInput() {
-        return micCheck = MediaUtil.isAvailableForVoiceInput(MainActivity.this);
+        return micCheck = MediaUtil.INSTANCE.isAvailableForVoiceInput(MainActivity.this);
```
amitiwary999: Avoid changes in the main activity. @chiragw15 is working on mainactivity. It may cause conflict.

mayank408: I have added that to the comments in the issue as said by @chiragw15 .

mariobehling:
@mayank408 What is the goal of converting everything? We have no urgency to do this and it just takes up developer time. The reviews on your PRs are very long and are taking up people's attention. From the user perspective there is no progress in the app.

mayank408:
@mariobehling We are converting the application from java to Kotlin with MVP. This will allow easy unit testing and as the crash reported by other users will also decrease drastically. This is under the issue #787 and #786

mariobehling:
@mayank408 Please only convert what is really necessary for a working app. Don't spend time on anything that is not really needed at this moment.

mariobehling:
@mayank408 Looking through more of the work that got done, it just seems like you guys are amazed about Kotlin. This is great for you, but right now it does not improve the app. 3+ developers have worked on the app since the program started and we don't have a usable outcome, but an app that constantly crashes even though the feature set is pretty small. The goal of the project is not "how many lines of code can be converted to Kotlin". The goal is to get an app that is usable as quickly as possible and iterate from there.

Issue #853, opened 2017-07-15
Use the correct language for text-to-speech


Orbiter:
**Actual Behaviour**

Currently TTS uses the default language as configured in the android device

**Expected Behaviour**

The TTS output must be made in the correct language.

How it works:

If the action object within susi answer contains a "language" key, use that language to use correct voice in text-to-speech. The action may still have no "language" object assigned, in that case still use the default language of the device.

In case that the required language is not available in the android device, a notification should be made that the language must be downloaded (or how it is done in android). While the required language is not present just use the default language.

**Steps to reproduce it**

get a susi respond which may look like
```
{
  "query": "hello",
  "count": 1,
  "client_id": "aG9zdF84NC4xODEuMTEyLjEzOA==",
  "query_date": "2017-07-15T21:31:09.833Z",
  "answers": [{
    "data": [{
      "0": "hello",
      "1": "",
      "token_original": "hello",
      "token_canonical": "hello",
      "token_categorized": "hello",
      "timezoneOffset": "-120",
      "language": "en"
    }],
    "metadata": {"count": 1},
    "actions": [{
      "type": "answer",
      "language": "en",
      "expression": "Hello!"
    }],
    "skills": ["/susi_server/conf/susi/en_0700_ai_play.json"]
  }],
  "answer_date": "2017-07-15T21:31:09.855Z",
  "answer_time": 22,
  "language": "en",
  "session": {"identity": {
    "type": "host",
    "name": "xx.xx.xx.xx",
    "anonymous": true
  }}
}
```

You may recognise the "language" object in the "actions" objects.
If you switch on TTS in android, then you must hear the answer in english language.

The language name is given in ISO-639-1.

chiragw15:
Work on this once #847 gets merged

PR #868, opened 2017-07-28
fix: Changing language in TTS

chiragw15:
Fixes issue #853  
Added language to TTS


mariobehling:
Please resolve conflicts.

chiragw15:
@chashmeetsingh @betterclever @amitiwary999 review please.

mariobehling:
How is this setting stored on the server?

chiragw15:
@mariobehling This is not a setting. The server provides a language attribute in response which represents the language of the answer. So, we have to you use it to specify TTS in which language it should give the output.

mariobehling:
Ah, yes. But, we also need to follow up on the internationalization question. Will do that separately.

Issue #877, opened 2017-07-31
issue with progress bar in login activity


amitiwary999:
**Actual Behaviour**

If user enters wrong credential then app shows alert dialog. pressing ok button close dialog but progress bar does not dismiss. So user can't do anything until close the app.

**Expected Behaviour**

progress bar should dismiss.

**Steps to reproduce it**

Enter wrong credential and press login button. Press ok button in alert dialog.

**Would you like to work on the issue?**

yes

PR #878, opened 2017-07-31
Fix: dismiss progress dialog in case of login error

amitiwary999:
Fixes issue #877

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
@@ -1,12 +1,15 @@
 package org.fossasia.susi.ai.login
 
 import android.app.ProgressDialog
+import android.content.Context
 import android.content.Intent
 import android.graphics.Color
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
```diff
@@ -90,9 +93,11 @@ class LoginActivity : AppCompatActivity(), ILoginView {
     }
 
     override fun onLoginError(title: String?, message: String?) {
+        progressDialog.dismiss()
```
chiragw15: @amitiwary999 Before calling onLoginError Method from presenter, progress dialog is dismissed here https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/login/LoginPresenter.kt#L98 . I am not able to reproduce the bug. See this https://youtu.be/JivszzGFh8s

amitiwary999: yes I have latest code. Yesterday I was facing this issue regularly. Closing it for now . will open it if I face this issue

PR #884, opened 2017-08-05
feat: Add feedback feature

### LINKED ISSUE #840: Adding Feedback support to Susi answers
**Actual Behaviour**

We don't have any feedback system as of now in Android

**Expected Behaviour**

A feedback system must be implemented similar to Susi Webchat as it is already implemented in Susi server.

Yes, will start working on this after converting MainActivity to MVP

chiragw15:
Fixes #840 

Changes: Add feedback feature.

.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -252,9 +252,9 @@ public class ChatFeedRecyclerAdapter extends RealmRecyclerViewAdapter<ChatMessag
         if (getData() != null && getData().isValid()) {
             if (index == getData().size()) {
                 if (isSusiTyping) {
-                    return new ChatMessage(-404, "", "", false, false, false, "", null, "", 0);
+                    return new ChatMessage(-404, "", "", false, false, false, "", null, "", "");
                 }
-                return new ChatMessage(-405, "", "", false, false, false, "", null, "", 0);
+                return new ChatMessage(-405, "", "", false, false, false, "", null, "", "");
             }
             return getData().get(index);
         }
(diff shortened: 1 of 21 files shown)


chiragw15:
@amitiwary999 @chashmeetsingh @betterclever please review

chashmeetsingh:
@chiragw15 Increase the font size of the time label. Rest seems good.

chiragw15:
@chashmeetsingh done. Ignore codacy issue for now. It is due to long parameterized method. Will have to change complete structure for it. Will do it in separate PR.

app/src/main/java/org/fossasia/susi/ai/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -252,9 +252,9 @@ public ChatMessage getItem(int index) {
         if (getData() != null && getData().isValid()) {
             if (index == getData().size()) {
                 if (isSusiTyping) {
-                    return new ChatMessage(-404, "", "", false, false, false, "", null, "", 0);
+                    return new ChatMessage(-404, "", "", false, false, false, "", null, "", "");
                 }
-                return new ChatMessage(-405, "", "", false, false, false, "", null, "", 0);
+                return new ChatMessage(-405, "", "", false, false, false, "", null, "", "");
```
betterclever: @chiragw15 What does -404 and -405 refer to here? It would be better if we use constants to refer them rather than numbers?

betterclever: Rest looks good :+1:

chiragw15: >What does -404 and -405 refer to here?

@betterclever these are two random IDs. They are never used. They are negative because we don't want them to clash with actual message IDs. These are used to refer empty( zero height) view holders. IMO, there is no need to define them as constants because they are never referenced again. We just wanted to put any random number, so we used these.

betterclever: Okay.

betterclever: Not absolutely necessary, but is better to write inline comment mentioning it in such peculiar cases so that it is easier for other contributors to understand the purpose.

mariobehling:
Please resolve conflicts.

mariobehling:
Thanks!

PR #886, opened 2017-08-06
Feat: add set query language option in android

### LINKED ISSUE #845: Option to set different languages
currently the android client sends the client language to the susi_server within the `language` attribute. I believe that this is taken from the default settings of android.

For testing, we need an option to overwrite the default setting with a given language.

Please add such a setting with the options:
* default language
* en
* de
* es
* fr
* it

amitiwary999:
Fixes #845 
1. add set query language option in android
2. Updated end point and settings keys to store settings on the server


amitiwary999:
@chiragw15 review

PR #889, opened 2017-08-09
Trainhotword

### LINKED ISSUE #806: Add feature to train hotword detection using Snowboy training API
As mentioned in this comment https://github.com/fossasia/susi_android/pull/710#issuecomment-312267662 , implement a feature to train hotword detection using snowboy training API.

Problem for now : 
The hotword detection uses a model file from snowboy website. This model file is of 2 types : `susi.pmdl` and `susi.umdl`. pmdl stands for personal model and umdl stands for universal model. The process of getting a personal model file (.pmdl) is simple. Just say susi thrice on snowboy website and download the personal model and use it. But the problem with personal model is that it is defined for a specific person. I am using my personal model (susi.pmdl) right now for hotword detection so it works great for me and people with similar voice like me but not for everyone. And to get a universal model file (.umdl), we need minimum 500 person to train the susi hotword on snowboy website by saying 'susi' thrice. Once we have a universal model file, the hotword will work great for everyone. But it may take time since right now I see only 10 people trained the model and we want 490 more. 

Alternate for this now : 
Snowboy provides an api to train and get a pmdl file. As mentioned in above comment, according to snowboy docs
```
You can define truly customized hotword for each of your end customer. Just ask them to say the hotword 3 times and a model will be trained on the fly!

    Endpoint: https://snowboy.kitt.ai/api/v1/train/
    Type: POST
    Return: a binary personal model (.pmdl), or error
```
We can generate a pmdl model for everyone by asking them say 'susi' thrice after installation of app. When the user will run the app after installation, he will be prompted to say susi thrice. These three recordings will then be sent to this http://docs.kitt.ai/snowboy/#api-v1-train API as a post parameter. The API will return a .pmdl file which then will be used for hotword detection. By this way everyone will use their personal model and hotword detection will work for them smoothly. Once we have universal model trained by 500 person, we can update the app and use universal model which will work for everyone and there won't be a need for every new user to train model with his/her voice. 

screenshot_2017-07-18-23-04-44

screenshot_2017-07-18-23-04-37

chiragw15:
Fixes #806 

Adds a fragment to train hotword using user's voice using following api. 
Endpoint: https://snowboy.kitt.ai/api/v1/train/
    Type: POST
    Return: a binary personal model (.pmdl), or error

The api takes in three audio files in .wav format and returns a .pmdl file which is used as a model for hotword detection. 

What I have done?

I was thinking to implement this feature in a way that when user say "susi" to train the model, the app should know what user has said. So, is user says something apart from "susi" to train the susi model with his voice, this should not work and ask user to train again. 

For that what I did was: I used google speech to text to take user's voice input and convert it into text. And then check if converted text is equal to "susi". If it is equal, then allow to proceed else repeat. 

But the problem in the above way is: To call the snowboy api, we have to pass 3 .wav audio files as parameter. Google Speech to text does not provide a way to get the audio files that is used to convert speech to text. So, the above way fails.

There is an another way by using Media recorder but in that, we will not be able to know when user started speaking and when it finished. Neither we will be able to check if user actually spoke "susi". 

If I try to use both simultaneously, Google's policy is that it allows microphone to take one input at a time. So, it will give an error while running both parallel to each other. 

I have tried various ways to do so. Even used ffmpeg android library to create a .wav file from local cache .amr audio file from google stt but it is not able to access the file.

This is the problem I am facing.
@amitiwary999 @betterclever @chashmeetsingh @dynamitechetan Do you guys have any other way to implement this?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,7 +26,7 @@ android {
 
     defaultConfig {
         applicationId "org.fossasia.susi.ai"
-        minSdkVersion 15
+        minSdkVersion 16
         targetSdkVersion 25
         versionCode 1
         versionName "1.0"
@@ -98,4 +98,6 @@ dependencies {
         transitive = true;
     }
     compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
+
+    compile 'com.writingminds:FFmpegAndroid:0.3.2'
 }
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -11,9 +11,6 @@ import android.content.pm.PackageManager
 import android.media.AudioManager
 import android.net.ConnectivityManager
 import android.os.*
-import android.speech.RecognitionListener
-import android.speech.RecognizerIntent
-import android.speech.SpeechRecognizer
 import android.speech.tts.TextToSpeech
 import android.speech.tts.UtteranceProgressListener
 import android.support.design.widget.Snackbar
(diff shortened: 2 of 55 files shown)


chiragw15:
Need some time for research. Will come back to this later after implementing other things.

mariobehling:
Description and screenshot missing in this PR.

mariobehling:
Please follow best practices!

mariobehling:
Another two weeks passed and it was still not possible to add a description.

PR #892, opened 2017-08-11
Feat: Store user setting in case of anonymous mode

### LINKED ISSUE #891: Save settings for Anonymous user
**Actual Behaviour**

Settings are not saved for the anonymous user. 

**Expected Behaviour**

Settings should be saved for anonymous user.

Yes

amitiwary999:
Fixes #891 

Changes: storing setting for the anonymous user using different sharedpreferences key.


amitiwary999:
@chiragw15 review

PR #900, opened 2017-08-16
Feat: make speech input full screen 

amitiwary999:
Fixes #856 

https://drive.google.com/file/d/0B8gIkH5cqXi3T2pOUk9fMTFTQWc/view?usp=sharing
problem is that we can't show partial text but it is similar like Siri.

I made another UI
https://drive.google.com/file/d/0B8gIkH5cqXi3dlMxcE1GUEFNZ0k/view?usp=sharing

PR #906, opened 2017-08-26
Revert "Feat: make speech input full screen "

chiragw15:
Reverts fossasia/susi_android#900

Okay so I found this library https://github.com/zagum/SpeechRecognitionView which is used to display the animation implemented in this PR. The code in this PR is exactly copied from the library which can cause trouble for us since it is apache licensed and we haven't given any credits to the author. Also, why is there a need to implement the library again in the app when you can just import it.  

@amitiwary999 please keep these things in mind while implementing things. You can not just simply remove license and copy the code. Please implement this using the library instead now.


amitiwary999:
it's my fault I tried library then to make it simple I tried code directly. And forgot to add a license. Sorry.

mariobehling:
Apache license is ok as it is compatible, but we should add this as a dependent library if needed.

chiragw15:
@amitiwary999 The preferred way would certainly be using library but if it is impossible to implement it, then add the code directly and give credits to the author.

amitiwary999:
I am trying again. Otherwise I will use code with license. Thanks

On Aug 26, 2017 10:08 PM, "Chirag Wadhera" <notifications@github.com> wrote:

@amitiwary999 <https://github.com/amitiwary999> The preferred way would
certainly be using library but if it is impossible to implement it, then
add the code directly and give credits to the author.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/fossasia/susi_android/pull/906#issuecomment-325144879>,
or mute the thread

PR #911, opened 2017-08-28
Fix: Added table action type.

### LINKED ISSUE #96: Add support for tables in Susi response

amitiwary999:
Fixes #96 
Retrieving data from the server (table action type response is different from other action types), save retrieved data using realm, retrieve saved data and UI design is done. The only part left is showing table type data from history when the user logged in.


chiragw15:
@amitiwary999 The problem in this UI is that it will fail when number of columns are greater than say 5.

mariobehling:
@amitiwary999 Please resolve merge conflicts and make the table movable if there are more columns.

chiragw15:
@amitiwary999 working on this one?

Issue #848, opened 2017-07-15
Use moshi to automap JSON to Kotlin class to reduce boiler plate code


chiragw15:
Use Moshi convertor factory for auto-mapping https://github.com/square/moshi. By this way, we can reduce 
```
class LocationResponse {

    @SerializedName("ip")
    var ip: String? = null

    @SerializedName("hostname")
    var hostname: String? = null

    @SerializedName("city")
    var city: String? = null

    @SerializedName("region")
    var region: String? = null

    @SerializedName("country")
    var country: String? = null

    @SerializedName("loc")
    var loc: String? = null

    @SerializedName("org")
    var org: String? = null

    @SerializedName("postal")
    var postal: String? = null
}
```
to
```
class LocationResponse (
    var ip: String?,
    var hostname: String?,
    var city: String?,
    var region: String?,
    var country: String?,
    var loc: String?,
    var org: String?,
    var postal: String?
)
```
By this way, we can even fit every susi response in single file, making overall 3 files i.e for susi response, for websearch and for location. See this tutorial for example. https://android.jlelse.eu/keddit-part-6-api-retrofit-kotlin-d309074af0 and this code https://github.com/juanchosaravia/KedditBySteps/commit/8295dfa18896a3ac18757e4c026fa1e7d5ec1600#diff-f419ad5f6c76e2617a08405003a66c76

mayank408:
@chiragw15 I will be working on this.

S2606:
@mayank408 Are you working on this?? or can I?

chiragw15:
@S2606 please proceed. Cheers. :+1:

PR #918, opened 2017-09-07
Add moshi converter for reducing boilerplate code

S2606:
Fixes https://github.com/fossasia/susi_android/issues/848


S2606:
@chiragw15 As per the previous conversation does it mean to convert all the files in susi folder into one file??

chiragw15:
Yes. rest->responses->susi contains all responses from Susi. Will be better if all the classes are written in a single file.

chiragw15:
And same goes for location type responses and web search responses

S2606:
location and responses are already done. Only susi is left to be done

S2606:
@chiragw15 Please review

chiragw15:
@S2606 please check why circle ci is not building

chiragw15:
@S2606 The build is failing. Please check. There are unresolved references of some Classes and variables.

S2606:
@chiragw15 Ok

S2606:
@chiragw15 I would like to talk about this PR.
As per the issue, each and every file needs to be converted to Kotlin for moshi but the thing is some files cannot be converted as they are indeed crashing the app.
So what i would suggest is to keep the same serialized form so that the app remain stable.

Do give your reply.

chiragw15:
@S2606 Ohh. Is there a specific reason that only some files are not converting. What is the major roadblock here? Is this a problem with moshi convertor? What alternative do we have? See, the main motive of the issue is to reduce the boiler plate code. If you have any other alternative which could reduce unnecessary 1000 lines to say 200 lines, that would be great. If not, it is still working fine currently.

S2606:
@chiragw15 The problem isnt with moshi converter. The problem is since some files cant be converted, what is the use of keeping both moshi as well as gson converter?? if gson will stay then might as well not change at all??For me frankly speaking bring moshi coverter isnt feasible.

chiragw15:
>the problem is since some files cant be converted

That's what I am asking. Why can't some files be converted?

>what is the use of keeping both moshi as well as gson converter?

No, we won't keep both. If moshi works fine, well and good else gson is doing it's job just fine.

>For me frankly speaking bring moshi coverter isnt feasible right now.

Any other option you have in mind. The main requirement is to reduce number of lines of unnecessary code.

S2606:
That's what I am asking. Why can't some files be converted?
Expected top level declaration. An error i seriously unable to solve it and there is not much regarding this on the internet.This error particularly comes when converting datum.java file into kotlin

chiragw15:
It's not necessary to convert all files to Kotlin. You can keep datum as it is and convert all other susi response files to one. Try this if it works.

S2606:
@chiragw15 I am trying it until now but it isnt working.

iamareebjamal:
Just adding to the discussion, GSON does not need `@SerializedName` if the property name matches the JSON field name

chiragw15:
But Moshi has more advantages than that. Nevermind. @S2606  Could you just remove the `@SerializedName` wherever possible?

S2606:
@chiragw15 Can i create a new PR??

chiragw15:
@S2606 yes please.

PR #922, opened 2017-09-10
Fixes #421: getSusiMessage is called sequentially by queries

### LINKED ISSUE #421: Order of messages not right
**Actual Behaviour**

When sent two or more continuous messages, their response order is not correct.

**Expected Behaviour**

The response from susi should be in correct order.

**Steps to reproduce it**

Send two consecutive messages without waiting for susi's reply.

**Screenshots of the issue**

screenshot_20161219-082024

**LogCat for the issue**

12-19 08:20:03.793 15122-15989/org.fossasia.susi.ai V/chirag: yo (123 ) Sent for response
12-19 08:20:04.537 15122-15988/org.fossasia.susi.ai V/chirag: added yo (123) to database
12-19 08:20:05.122 15122-16014/org.fossasia.susi.ai V/chirag: hi (124 ) Sent for response
12-19 08:20:05.125 15122-16013/org.fossasia.susi.ai V/chirag: added hi (124) to database
12-19 08:20:05.632 15122-15122/org.fossasia.susi.ai V/chirag: Got response of hi (124)
12-19 08:20:05.659 15122-16026/org.fossasia.susi.ai V/chirag: added Hi, I'm Susi(125) to database
12-19 08:20:07.166 15122-15122/org.fossasia.susi.ai V/chirag: Got response of yo (123)
12-19 08:20:07.195 15122-16053/org.fossasia.susi.ai V/chirag: added Yo this Instagram update is crazy. You can livestream!(126) to database

Yes

dabbler011:
Fixes #421 

onComputeMethod() cannot be called directly from computThread class if waiting for response instead other messages will be send to server after onSuccess or onFailure of previous response.

@chiragw15 please review if this works.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
@@ -21,6 +21,7 @@ import org.fossasia.susi.ai.rest.responses.susi.SusiResponse
 import retrofit2.Response
 
 import java.util.*
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Presentation Layer for Chat View.
@@ -47,7 +48,7 @@ class ChatPresenter(chatActivity: ChatActivity): IChatPresenter, IChatModel.OnRe
     var check = false
     var atHome = true
     var backPressedOnce = false
-    var queueExecuting = false
+    @Volatile var queueExecuting = AtomicBoolean(false)
 
     override fun onAttach(chatView: IChatView) {
         this.chatView = chatView
@@ -294,7 +295,7 @@ class ChatPresenter(chatActivity: ChatActivity): IChatPresenter, IChatModel.OnRe
 
     private inner class computeThread : Thread() {
         override fun run() {
-            if(!queueExecuting) {
+            if(queueExecuting.compareAndSet(false,true)) {
                 computeOtherMessage()
             }
         }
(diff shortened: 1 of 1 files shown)


chiragw15:
>onComputeMethod() cannot be called directly from computThread class if waiting for response instead other messages will be send to server after onSuccess or onFailure of previous response.

That is not the case. Whenever a message is sent to the sever, the network call is made on a separate thread. So, the app does not have to wait for the response of previous message to send a new message.  What you have done right now is: You have set a boolean flag `queueExecuting` in a certain thread and then accessing the value of that boolean variable from other thread, which is wrong. Since both threads run in parallel, you can not be sure that the variable has been assigned the right value before you are using it.

dabbler011:
@chiragw15 for that if queueExecuting is volatile and AtomicBoolean class is used fine syncronization is being achieved across threads also.
Shall i send PR with same changes?

chiragw15:
@dabbler011 But what is the need of doing that. The network calls are asynchronous right now. The response of second query is not depended on the first. Is the current code doing something else?

dabbler011:
commited the changes i commented about. it will now send only one message at a time to server.

dabbler011:
@chiragw15 please review if it is fine.
it allows messages to send and recieve on single thread so order becomes right.

chiragw15:
@dabbler011 I pulled the code from this PR to test. Works fine. Have you implemented something else in the APK you referred above?

dabbler011:
no, apk was for the same code.

chiragw15:
Cool. The build will be fixed after merging of #929 . The PR can be merged then.

mariobehling:
Ok, merging. Let's see if it is working.

Issue #613, opened 2017-05-13
Increase speed of app wherever possible


chiragw15:
**Actual Behaviour**

The response of app is slower even if the response from server is fast? I don't know the reason yet. Maybe poor implementation or a need of better algorithm. Need to figure it out. 

**Expected Behaviour**

If the response from server is fast, the response from app should be fast too.

**Mini-Project**
Increase the speed of app wherever possible.

@amitiwary999 If you find any poor implementation in app which can be improved, comment on this issue.

amitiwary999:
@mariobehling  We can use google cloud speech api. It return text result in real time just like google assistant. @chiragw15 I am trying to implement it in a sample app by following there docs it will be better if you also look into it.

mariobehling:
@amitiwary999 Please open a dedicated issue.

amitiwary999:
@chiragw15 your implementation looks good . I am opening a issue for it you can make a pull request. I too trying to find some other way to make susi faster.

amitiwary999:
@chiragw15 currently some function like calling someone,sending email, setting alarm etc. which not depend on susi response are kept under susi response in mainactivity.  If we are planing to keep these functions hardcode then we should keep these functions outside of it.If you agree then I will open a issue. @mariobehling @Orbiter needed your suggestion.

chiragw15:
@mariobehling I have removed all harcoded features. They were interfering with actual susi response. Also, it makes the app a little faster, efficient and reduces app size. The code is in my fork https://github.com/chiragw15/susi_android/tree/removehard . To test it here is an [apk](https://drive.google.com/file/d/0B6IO4qTKhVUCRHJDYzEydElSamc/view?usp=sharing). If you feel it has increased the speed of the app, I'll create a PR.

mayank408:
Merged

chiragw15:
@mayank408 This issue is a parent issue. We have to continuously check for implementation and better techniques for increasing speed of app. Only a part of it is solved.

arundhati24:
The app has a new UI flow now. The skills loaded very slowly due to a number of API calls at almost the same time. The skills load faster now as the API calls are now being made optimally. Hence, closing this issue.

PR #954, opened 2017-09-21
fix: Skills loading speed increased

collinx:
This PR is in reference to [#613](https://github.com/fossasia/susi_android/issues/613). 

Changes: 
Earlier client builder was being built at every instance of skill listing call. Now, Client Builder is being initialized only once and the same object is used in each skill call. This has reduced the skill loading time to half.


collinx:
@chiragw15 review the PR.

collinx:
thanks :thumbsup: @chiragw15

PR #957, opened 2017-09-22
Updated Gradle file

### LINKED ISSUE #956: Unresolved Reference kotlinx and Updated Gradle
**Actual Behaviour**

Android Studio shows unresolved reference for kotlinx.

**Expected Behaviour**

There should be no warning for kotlin support in android studio.

**Steps to reproduce it**

Build the cloned repo

Yes

yashLadha:
- Fixed unresolved reference for kotlinx
- Updated gradle version for instant run
Fixes #956 

Changes: Updated gradle version and applied plugin for kotlin

Screenshots for the change: 
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.2'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.4.1'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -28,3 +28,4 @@ allprojects {
     }
 }
 
+apply plugin: 'kotlin-android-extensions'
```
collinx: Kotlin plugin is already applied in build.gradle (app module).

yashLadha: But is giving warnings when compiled. Giving unresolved reference error for kotlinx. After adding this all the warnings are gone.

collinx: There are no such warnings on the compilation on my side.  Do check again.

yashLadha: Yaa, may be it is not more, after installing the compiler.

PR #969, opened 2017-09-24
Test gem integration

iamareebjamal:
Test gem availability for Fastlane Integration in CircleCI


iamareebjamal:
Gem is available and configured correctly

PR #974, opened 2017-09-24
fix: Don't display empty skillgroups

### LINKED ISSUE #968: Susi skill groups with no skills have no message
**Actual Behaviour**

The skill listing page displays skill groups that are empty, with no message.

**Expected Behaviour**

It should either not list the skill group or display that there are no skills found, like the [skills.susi.ai](https://skills.susi.ai) website does.

**Steps to reproduce it**

Open the SUSI.AI Skills page

**Screenshots of the issue**

 
Here, Lifestyle, Health & Fitness, and Social don't have any skills.

Yes

Muks14x:
Hide empty skillgroups

Fixes #968 

Changes: Hid the skill-group title and recyclerview if the group had no skills under it.

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
@@ -52,8 +52,11 @@ class SkillListingPresenter: ISkillListingPresenter,
     override fun onSkillFetchSuccess(response: Response<ListSkillsResponse>, group: String) {
         skillListingView?.visibilityProgressBar(false)
         if (response.isSuccessful && response.body() != null) {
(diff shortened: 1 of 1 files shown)


chiragw15:
@Muks14x alternatively you can check [here](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt#L55) is skillMap is empty, then don't add to arraylist. Which way would be better?

Muks14x:
I tried that. Works, of course - code would look a lot cleaner too 😅. The only positive in this is if we decide to show 'No skills found' instead later on. Should I switch?

chiragw15:
@Muks14x switch. This is a continuous list of lists. If we show "No skills found" for empty groups it will look odd imo. Better would be to not show anything.

PR #988, opened 2017-09-26
Enhanced susi skill frontend

### LINKED ISSUE #987: Enhanced Susi Skills Activity
Enhancing the frontend fo susi skills activity by using ripple effect and material fonts as described in material guidelines https://material.io/guidelines/style/typography.html

yashLadha:
Fixes #987 

Changes: Added Roboto font following material guidelines. 

Screenshots for the change: 

*Ripple effect*
screenshot_1506413431

*Font Changes*
Before

screenshot_2017-09-26-13-42-50

After

screenshot_2017-09-26-13-44-14

**NOTE: The changes of screenshot of font changes are from Samsung galaxy s3 neo running android 5.1.1**

### DIFF
--- a/app/src/main/res/layout/item_skill.xml
+++ b/app/src/main/res/layout/item_skill.xml
@@ -10,6 +10,7 @@
         android:layout_width="@dimen/websearch_card_width"
         android:layout_height="wrap_content"
         android:layout_marginBottom="@dimen/message_card_margin_small"
+        android:foreground="?attr/selectableItemBackground"
         android:layout_marginLeft="@dimen/margin_card"
         android:layout_marginTop="@dimen/message_card_margin_small"
         app:cardBackgroundColor="@color/susi_message_bg">


chiragw15:
@yashLadha This does not make much difference imo. @iamareebjamal opinion?

iamareebjamal:
Though ripple effect is a nice addition. Bundling fonts which should be natively available on system is not an option. This adds nothing much and bloats the apk

Roboto is installed as default font on Android and if an OEM is trying to replace it, we shouldn't circumvent it. It is not our job to make them apply Material Design. Also, this may break user installed theme which overrides a font, and create a bad user experience. We should let the system handle the fonts

If so crucial, please use this approach instead https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts.html

yashLadha:
So should i remove the font changes @iamareebjamal

chiragw15:
@yashLadha yes. Leave the fonts as it is. Just add ripple effect.

PR #997, opened 2017-09-28
fix: Backbutton press in SkillsActivity

### LINKED ISSUE #995: Bug in handling back button press from About and Settings fragments
**Actual Behaviour**

When the back button is pressed from the About and Settings fragment **after being launched from SkillDetails fragment**, causes the title to change to "SUSI.AI Skills", with the same fragment being displayed.

**Expected Behaviour**

The back button should close the current fragment and show the previous fragment on the stack.

**Steps to reproduce it**

Open the skill details activity for some skill (eg, Cooking Guide). Now, open the About or settings page. On pressing back once, the title changes to "SUSI.AI Skills", without the fragment closing.

**Screenshots of the issue**

  

Yes, I would.

Muks14x:
Fixes #995 

Changes: 

AboutUsFragment and ChatSettingsFragment both inherited from android.support.v4.app.Fragment, whereas SkillDetailsFragment and SkillListingFragment use android.app.Fragment. This was causing problems with the fragment manager stack. So, for uniformity, changed everything to use the compatibility library.

Now, on pressing back, everything goes back to the skill listing fragment.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/SkillsActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/SkillsActivity.kt
@@ -32,8 +32,9 @@ class SkillsActivity : AppCompatActivity() {
         setContentView(R.layout.activity_skills)
 
         val skillFragment = SkillListingFragment()
-        fragmentManager.beginTransaction()
+        supportFragmentManager.beginTransaction()
                 .add(R.id.fragment_container, skillFragment, TAG_SKILLS_FRAGMENT)
+                .addToBackStack(TAG_SKILLS_FRAGMENT)
                 .commit()
     }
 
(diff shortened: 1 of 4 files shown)


Muks14x:
Yes, one would have to press back twice to go back, but opening up whatever was directly above in the hierarchy is what feels most natural to me in android. Instead of jumping back 2 screens, don't we generally expect the last screen we saw to be displayed, on going back? O.o

collinx:
Imo, settings and about us fragments should direct the user to chat activity on back button pressing. what do you think @chiragw15 ?

Muks14x:
It's not a huge change😄. I'll change it that way if needed. It's pretty easy to change it back too.

chiragw15:
Since the about and settings fragment are opened from clicking on the item in the menu of Skill Activity, changing activity on pressing back should not be done imo. So, implement something like, when user is on Settings, About or Skill Detail fragment and he presses back button, open skill List fragment and when on skill List fragment open Chat Activity. What say @collinx @Muks14x

collinx:
@chiragw15 fine by me 👍

Muks14x:
@chiragw15 If I understand you correctly, I think that's how I've implemented it, can you check it out and see if you need changes?

PR #1027, opened 2017-10-13
fix: Do not concatenate strings in loop

trubitsyn:
This PR is reference to #613.

Changes: Use builtin joinToString method instead of concatenating strings in loop at https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt#L130


trubitsyn:
Please note that the fixed implementation does not append space to the end of string. If having a space in the end is critical, please tell me and I'll change the commit.

PR #1031, opened 2017-10-19
Fix permissions not granted for some UI tests.

### LINKED ISSUE #1015: Fix flaky UI tests
```
java.lang.RuntimeException: Waited for the root of the view hierarchy to have window focus and not be requesting layout for over 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus. Otherwise, something is seriously wrong. Selected Root:
```

There are flaky UI tests in the app, either these should be fixed or removed (replaced by RoboElectric unit tests) since they fail the build on occasional basis making it hard to validate the status of repository

Tapchicoma:
Signed-off-by: Yahor Berdnikau <egorr.berd@gmail.com>

Fixes #1015 

Changes: grant permission for UI tests using [GrantPermissionRule](https://developer.android.com/reference/android/support/test/rule/GrantPermissionRule.html)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -86,9 +86,9 @@ dependencies {
 
     //unit test
     androidTestCompile "com.android.support:support-annotations:$support_lib_version"
-    androidTestCompile 'com.android.support.test:runner:0.5'
-    androidTestCompile 'com.android.support.test:rules:0.5'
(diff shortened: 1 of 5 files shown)


Tapchicoma:
Seems to be not related to the linked issue as tests are running on emulator with android 22.
But still worth to have it.

PR #1055, opened 2017-11-22
Display login email in settings

### LINKED ISSUE #1054: Displaying the email of the logged-in user
The email of the logged-in user should be displayed in Settings. Currently, the identity of the user is not  displayed anywhere after login.

Yes

Nirvan101:
Fixes #1054   

Changes: Added code to ChatSettingsFragment.kt and pref_settings.xml.  I have added a new item in Settings . It displays the email with which the user is currently logged in. In case the user is not logged in, it displays 'Not logged in' .

This will keep the user informed about which account he is logged in with.

Screenshots for the change: 

one
two


DipanshKhandelwal:
Hey , @Nirvan101 .
I guess it would be better if we follow template something like-  [Username](email).

PR #1109, opened 2018-02-14
Made background data invisible while skill fetch error message is displayed (Fixed #1106)

### LINKED ISSUE #1106: Skill fetch error message overlaps skill groups
**Actual Behaviour**

Currently the error message that is being displayed in case of an error while fetching skills overlaps the default skill groups. This creates a mess on the skills screen and as a result the error message is not quite clear to the user.

**Screenshot for current behaviour :** 

error_skill_fetch_1

**Expected Behaviour**

The skill groups should be invisible while the error message is being displayed.

**Screenshot for expected behaviour :** 

error_skill_fetch_22

Yes

arundhati24:
Added necessary code to make the skill groups invisible while the
skill fetch error message is being displayed. This makes the error
message clear to the user and also makes the screen look cleaner.

Fixes #1106  

Screenshots for the change: 

Before : 

error_skill_fetch_1

After : 

error_skill_fetch_22


arundhati24:
@chiragw15  Please review this pull request.

PR #1141, opened 2018-02-28
Fix #1126-Edittext issue in the login page fixed.

### LINKED ISSUE #1126: Edittext issue in the login page.
**Actual Behaviour**

When the login screen is launched the enter email edittext gets the focus, but the soft-keyboard does not pops up automatically instead the user has to tap on it for the soft-keyboard to appear.

**Expected Behaviour**

When the enter email edittext gets the focus, the soft-keyboard should pop up automatically.

**Steps to reproduce it**

Launch the app.

Yes.

sauravvishal8797:
Fixes #1126 
Changes: Soft-keyboard automatically pops up when the Edittext to enter email receives focus.

### DIFF
--- a/app/src/debug/AndroidManifest.xml
+++ b/app/src/debug/AndroidManifest.xml
@@ -43,7 +43,7 @@
         <activity
             android:name=".login.LoginActivity"
             android:theme="@style/Login_Theme"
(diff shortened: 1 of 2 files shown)


sauravvishal8797:
@chiragw15 Please review.

Issue #1127, opened 2018-02-22
Edittext issue in the reset password page.


sauravvishal8797:
**Actual Behaviour**

On opening the reset password page, the enter email edittext automatically gets focus but soft-keyboard does not pops up automatically.

**Expected Behaviour**

When the enter email edittext gets focus, the soft-keyboard should pop up automatically.

**Steps to reproduce it**

Go to the forget password page from the login activity.

**Would you like to work on the issue?**

Yes.

PR #1138, opened 2018-02-26
Fix #1127-Reset password page edittext issue resolved.

sauravvishal8797:
Fixes #1127 

Changes: Soft-keyboard automatically pops up when the enter email Edittext gets focus.

PR #1144, opened 2018-03-01
#1127 Problem Resolved

developershivamshukla:
Edittext issue in the reset password page. #1127 resolved
Now Soft Keyboard are popup when activity changes from login to forgot password.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -14,6 +14,7 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
     <application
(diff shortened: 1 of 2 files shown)


sauravvishal8797:
@developershivamshukla  Please close this PR(Always do check before opening a PR whether a PR is already opened for the issue or not). I clearly mentioned in the issue that I'd be working on it and have already opened a PR for the same. #1138 Therefore please see that from next time you consult the person working on the issue whether he/she wants to give it to someone else or not, before working on the issue yourself as it will just waste your and the author's time and effort. Instead you can create new issues if you find any and can work on that.

Issue #1080, opened 2017-12-18
Pressing back button after Settings->Login closes the app 


Nirvan101:
**Actual Behaviour**
If the user is not logged in , going to Settings->Login opens the Login page. Here if the user presses the back button, the app closes instead of going back to the settings page. 
 
**Expected Behaviour**
The app closes on pressing the back button in LoginActivity. This should not happen if the user has come to the Login page from Settings->Login.
    If the user has come to LoginActivity from Settings->Login, pressing the back button should take the user back to Settings page.

**Steps to reproduce it**
Open the app without logging in. Go to Settings and press Login. The Login page opens. Press back button. The app will close instead of going back to the Login page.

**Would you like to work on the issue?**
Yes

ribhav-pahuja:
@Nirvan101 are you still working on this?

Nirvan101:
No. Feel free to take it up!

anipmehta:
@Nirvan101 I just submitted a PR for this issue and if you think this is not able to solve the issue then you can reopen this.

ShubhamAvasthi:
I would like to work on this issue.

Nirvan101:
Sure, though I must warn you. It seems like a number of people have tried and failed on this quest before you. I remember it being a pretty difficult problem and I couldn't see any straightforward way to do this with the existing design.

ShubhamAvasthi:
> 
> 
> @ShubhamAvasthi Sure, though I must warn you. It seems like a number of people have tried and failed on this quest before you XD. I remember it being a pretty difficult problem and I couldn't see any straightforward way to do this with the existing design.

I understand and will atleast try. What was the issue in this pull request #1116 ?

atm1504:
@mariobehling @iamareebjamal please close this issue. This issue has already been resolved.

PR #1151, opened 2018-03-02
Fixes #1149 -Added confirmation before exit on the login screen

### LINKED ISSUE #1149: Add an Alert dialog to confirm before closing the app when (logging in) in the app.
**Actual Behavior**

On exiting the application at the time of login, the application get close directly , may be user don't want to close it & mistakenly it get closed. 

**Expected Behavior**

A popup must appear that confirms user ,that he/she really wants to close the application or not.

**Step to reproduce**

1. Reinstall or clear the data of the app.
2. Launch the app.
3. Press the back button when the login screen appear.

**Would you like to work on the issue**?

Yes.

ujjwalagr:
**Fixes** #1149 -Added confirmation before exit.

**Changes:**
 1. Updated LoginActivity.kt  overriding the back button pressed function and added confirmation before exit when user is (logging in) in the app.
 2. Added Toast that confirm the exit status at the login activity

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/login/LoginActivity.kt
@@ -4,6 +4,8 @@ import android.app.ProgressDialog
 import android.content.Intent
 import android.graphics.Color
 import android.os.Bundle
+import android.os.Handler
+import android.support.v7.app.AlertDialog
 import android.support.v7.app.AppCompatActivity
 import android.view.View
 import android.view.inputmethod.EditorInfo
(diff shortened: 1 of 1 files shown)


ujjwalagr:
@chiragw15 please review.

arundhati24:
@ujjwalagr 
As you have not added the screenshot for the change, as far as I understand, you want to prompt the user if the user happens to press the back button that will cause her/him to exit from the app. If that is true, I would like to point out that the app already prompts the user for the same by displaying a toast that asks the user to press the back button twice in case s/he wants to exit from the app. So, I guess this extra confirmation is not required.

ujjwalagr:
@arundhati24 I already mentioned that user is not getting any confirmation on the login screen, yeah you are right that afterwards on the Chat screen user gets a toast but I found the issue on the login screen(LoginActivity.kt).

arundhati24:
@ujjwalagr  Actually issue #1080  addresses a similar issue but when the user accesses the login screen from the settings. So, I suggest you to make the title of the issue a bit clearer. Also, it would be better if you could add a little clearer description regarding the change you made.

ujjwalagr:
Previously
whatsapp-video-2018-03-02-at-10 01 57-pm

After updation
whatsapp-video-2018-03-02-at-9 57 30-pm

ujjwalagr:
@arundhati24 Please review.

arundhati24:
@ujjwalagr I would suggest you to keep the toast message same as that used in the main app.

ujjwalagr:
@arundhati24 Okay ,So do i have to send the pr again?

arundhati24:
@ujjwalagr No. You can make the required changes and then you can squash those commits into one.

ujjwalagr:
@arundhati24 Squashing would going to be complicated rather I close this request And create a new PR simply.Is it okay?

PR #1169, opened 2018-03-10
Retain scroll position of chat feed on configuration change

### LINKED ISSUE #1164: Scroll state of the list not preserved
**Actual Behaviour**

Scroll state of the conversation list is not retained on rotating the device from portrait to landscape mode.

**Expected Behaviour**

Scroll state of the list must be preserved.

Yes

harshitbansal05:
Fixes #1164 

Currently, the state of the list is not conserved when the device configuration changes from portrait to landscape. It was due to the logic in `OnLayoutChangeListener`. This pull request solves the issue by taking into consideration the configuration change of the device.


harshitbansal05:
@chiragw15 please review..

PR #1170, opened 2018-03-10
Removed confirmation before exit (Fixed #1166)

### LINKED ISSUE #1166: Remove confirmation before exit
**Actual Behaviour**

Currently, the app displays a toast to prompt the user, in case the user hits the back button.

**Expected Behaviour**

The app should not display any toast or dialog to confirm exit, unless there is some task running in the background (say, download), as this breaks user experience. Moreover, none of the standard apps implement this.

Yes

arundhati24:
Fixes #1166 

Changes: Removed the toast message that was being displayed to prompt the user before exiting from the app as that might break user experience.


arundhati24:
@chiragw15 Would you please review this PR? :)

PR #1186, opened 2018-03-15
Using PrefManager class instead of creating separating Preference object

### LINKED ISSUE #1083: Use Predefined PrefManager rather than creating new SharedPref's
**Actual Behaviour**

we are creating new SharedPref's in `WelcomeActivity` class.

**Expected Behaviour**

We can use our predefined `PrefManager` class to perform that task

**Steps to reproduce it**

Use `PrefManager` class

**Screenshots of the issue**

image

Yes

mamakancha:
Fixes #1083 

Changes: Using PrefManager class instead of creating separating Preference object


mamakancha:
@chiragw15 Please review and merge

PR #1202, opened 2018-03-18
Fix #1201: Addition of a delete chat history option in the app

### LINKED ISSUE #1201: Additon of delete chat history option in the app
**Actual Behaviour**

Presently, there is no chat history delete option present in the app.

**Expected Behaviour**

There should be a chat history delete option present which can help the user delete the whole chat history and help clear the unwanted messages for the user.

**Steps to reproduce it**

I can use the delete messages feature of realm database so as to clear the chat history.

Yes

aniket-agarwal1999:
Fixes #1201 

Changes: 
I have added a delete chat history option in the settings fragment of the app. This will help user delete his/her chat history on their wish.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
+++ b/app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
@@ -21,6 +21,7 @@ public class Constant {
     public static final String HOTWORD_DETECTION = "hotword_detection";
     public static final String SHARE = "share";
     public static final String LOGIN_LOGOUT = "login_logout";
+    public static final String DELETE_CHAT = "delete_chat";
     public static final String RESET_PASSWORD = "reset_password";
     public static final String CHANGE_PASSWORD = "Change Password";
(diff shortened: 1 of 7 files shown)


aniket-agarwal1999:
@arundhati24 I will correct today only

aniket-agarwal1999:
@arundhati24 Actually I have not just started the chat activity. Actually I have cleared the data in the Realm database which was stored locally. However when user logs out and log in again actually the data is fetched from the server side. I will definitely try to delete the data even from the server side, but yes the deleting data from the local device is working just fine.

batbrain7:
@aniket-agarwal1999 how will you delete the data from the server side from the app?

aniket-agarwal1999:
@batbrain7 Yes that is the problem I am facing

chiragw15:
@aniket-agarwal1999 create an api on the server to delete the user chat history and call it from the app when user clicks on "delete all messages" option.

PR #1205, opened 2018-03-18
Fix #1203 : No Internet Connectivity dialog is shown on pressing skip

### LINKED ISSUE #1203: Issue on pressing skip when no internet connectivity.
**Actual Behaviour**

Presently if there is no Internet connection then pressing login button No Internet Connectivity dialog box is shown but when you press skip, chat activity is opened.

**Expected Behaviour**

If there is no Internet connectivity then No Internet Connectivity dialog box should be shown instead of opening the chat activity.

**Steps to reproduce it**

Press skip button when there is no Internet connection in your mobile phone.

Yes.

raj10071997:
Fixes #1203
Changes: No Internet Connectivity dialog is shown on pressing skip when there is no Internet connection in the phone instead of opening the chat activity.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/login/LoginPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/login/LoginPresenter.kt
@@ -49,9 +49,14 @@ class LoginPresenter(loginActivity: LoginActivity): ILoginPresenter, ILoginModel
     }
 
     override fun skipLogin() {
-        utilModel.clearToken()
-        utilModel.saveAnonymity(true)
-        loginView?.skipLogin()
+        if(NetworkUtils.isNetworkConnected()) {
+            utilModel.clearToken()
+            utilModel.saveAnonymity(true)
+            loginView?.skipLogin()
+        } else {
+            loginView?.onLoginError(utilModel.getString(R.string.error_internet_connectivity),
+                    utilModel.getString(R.string.no_internet_connection))
+        }
     }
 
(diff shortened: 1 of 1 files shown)


arundhati24:
@raj10071997 IMHO, this is not an issue. This is the expected behaviour of the app. If you are not connected to the internet then obviously you cannot log into your account because you cannot connect to the server. Hence, when you fill in your credentials and tap the Log In button, you see a dialog that tells you that there's no internet connection. But, when you press skip button, you do not require any authentication from the server, hence you are able to enter the Chat Activity without a dialog telling you about no internet connectivity.

raj10071997:
@arundhati24  If there is no Internet then what is the use of directing user to chat activity, if he/she can't even use the application. IMO, the skip button works exactly the way it should be. I mean it skips the authentication process for the user but when there is no Internet connection, there is no meaning of directing user to chat activity. And even if we direct user to chat activity, the user can't chat with Susi, can't see the skills and use them, and can't even see the effects of changing the settings. So without Internet connection, the user can't do anything and that's why I though that there should be a dialog when there is no Internet connection on pressing skip just like the dialog box when we press login button or sign up button

iamareebjamal:
This is not a standard behaviour across any app. Handling no connection is important. Maybe show this in ChatActivity or just let the user send the messages and realize that there is no internet connection. I am closing this for now, but we can have a discussion and will reopen if other maintainers think this is the desired behaviour

PR #1229, opened 2018-05-11
"double circularProgressBarFix"

### LINKED ISSUE #1227: Two Circular progress Bar onSwipetorefresh() in skill activity
#**Actual Behaviour**
In SUSI.AI Skills activity when we swipe to refresh the skills data along with one default circular progress bar,  Another progress bar which we defined in layout also starts. 

**Expected Behaviour**
if we hit Swipe to refresh only default circular progress 
bar should be shown. when the activity is loaded then only progress bar from activity layout should be shown

**Steps to reproduce it**
when we are loading groups in "fetchGroups()" in SKillListingPresenter() we should check for we are loading on activity creation or for onRefresh

**Screenshots of the issue**
 

Yes

storytellerr:
Fixes #1227 

Changes: added a check before fetchGroups() in skillListingPresenter that we are loading skill data on activity creation or for Refresh

Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
@@ -28,7 +28,7 @@ class SkillListingPresenter: ISkillListingPresenter,
         this.skillListingView = skillListingView
     }
 
-    override fun getGroups(swipeToRefreshActive:Boolean) {
+    override fun getGroups(swipeToRefreshActive: Boolean) {
         skillListingView?.visibilityProgressBar(true)
         skillListingModel.fetchGroups(this)
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
```diff
@@ -28,7 +28,7 @@ class SkillListingPresenter: ISkillListingPresenter,
         this.skillListingView = skillListingView
     }
 
-    override fun getGroups() {
+    override fun getGroups(swipeToRefreshActive:Boolean) {
```
iamareebjamal: Space after :

app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/contract/ISkillListingPresenter.kt
```diff
@@ -6,6 +6,6 @@ package org.fossasia.susi.ai.skills.skilllisting.contract
  */
 interface ISkillListingPresenter {
     fun onAttach(skillListingView: ISkillListingView)
-    fun getGroups()
+    fun getGroups(swipeToRefreshActive:Boolean)
```
iamareebjamal: Space after :

PR #1239, opened 2018-05-16
Fix failing UI tests

### LINKED ISSUE #1238: Fix Failing UI tests
**Actual Behaviour**

Currently, the UI tests are failing because the permissions required while running the UI tests are not being granted properly.

**Expected Behaviour**

The required permissions should be granted when the dialog pop ups while running the UI tests on a device/emulator. This happens for API level >= 23 because from here the permissions are requested by the app when needed.

Yes

arundhati24:
Trying to fix #1238

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.2'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.2'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.4.1'
         classpath "io.realm:realm-gradle-plugin:3.5.0"
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Tue Dec 12 15:49:37 IST 2017
+#Tue May 15 13:50:43 IST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,7 +26,6 @@ repositories {
 
 android {
     compileSdkVersion rootConfiguration.compileSdkVersion
-    buildToolsVersion rootConfiguration.buildToolsVersion
 
     defaultConfig {
         applicationId "ai.susi"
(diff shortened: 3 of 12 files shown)


arundhati24:
@batbrain7 Could you please have a look at the issues and suggest changes. Please have a look at ChatActvityTest.kt where I have tried a different approach to use ui automator to handle clicks to grant permsission.
The GrantPermissionRule doesn't seemt to work properly. Also, it is in beta as of now.

batbrain7:
@arundhati24 are the permissions being enabled by this on each test?

arundhati24:
@batbrain7 When we run the ChatActivityTest.kt (I ran it on my device Redmi Note 4) what we observe is a dialog box asking for Device Location access. But, nothing happens after that. So I dont think these are being granted. But yes, we can manually click on the allow button while running tests. What we need is the ui automator to click the ALLOW option on the dialog so that we can grant all 3 permissions and move on to the Chat Activity for running further tests.

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.2'
```
batbrain7: leave the gradle properties as is

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Dec 12 15:49:37 IST 2017
+#Tue May 15 13:50:43 IST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
batbrain7: same here please don't commit gradle properties

iamareebjamal: These should be upgraded to 4.6 but in a different PR

arundhati24:
@batbrain7 I am sorry, will fix this soon. By mistake I committed those files.

gradle/wrapper/gradle-wrapper.properties
arundhati24: @iamareebjamal sure.

iamareebjamal:
Also, implementing UI tests is a hard task for apps which have no constant designs and changing specifications. Since we are using MVP, if we move all logic to presenter components, there won't be a need of most UI tests at all. This problem with UI tests will only increase in future. So we should move in direction of increasing unit tests so that they are both easy, small and quick to code and run and we have less dependence on flaky UI tests

If the view is implementing small sections of display logic, just verifying that the presenter calls those is enough to guarantee that the UI is going to work. This feature makes testing easier in MVP than MVVM, however MVVM is more decoupled and less prone to memory leaks

iamareebjamal:
For reference on how to create extensive unit tests in MVP, check bar code module tests in Orga App, each logic, including permissions in encapsulated in order to test by mocking and abstraction

iamareebjamal:
I think this should be closed @arundhati24

arundhati24:
@iamareebjamal  Yes. This should be closed now.

PR #1241, opened 2018-05-16
Added message if no skill is found

### LINKED ISSUE #1240: No message given if the skill is not found
**Actual Behaviour**

Whenever we search the susi skill in SkillActivity.kt, if there is no skill found then nothing happens

**Expected Behaviour**

Ideally we must be shown a message saying "No skill found"

**Steps to reproduce it**

Search any skill that is not listed in the SkillActivity.kt.

Yes

batbrain7:
Fixes #1240 

Changes: Added a toast message if no skill is found in the app on searching


batbrain7:
@arundhati24 @chiragw15 review and merge

PR #1248, opened 2018-05-16
Removed menu from About Us Fragment (Fixed #1242)

### LINKED ISSUE #1242: Remove menu from About Activity
**Actual Behaviour**

Currently, there is a menu in the About activity, which contains only one option i.e. Settings. This is not a correct UI flow because we reached the About activity from the Skills Activity menu, which has the options About as well as Settings. Now, we allow the user to reach the Settings Activity from the About activity. Also, once we reach Settings activity from About, the back button should direct us to About activity, which is not the case as of now.
Consider a situation, when we might have more number of options in our menu. Then, also it would not be correct to have a sub-menu like this in each activity.

**Expected Behaviour**

The About activity doesn't need a menu. If the user wishes to open Settings Activity, s/he should press back button to return to the SkillsActivity from where s/he can choose the Settings option from the menu.

**Screenshots of the issue**

settings_in_about

Yes

arundhati24:
Fixes #1242 

Changes : Now the About Us Fragment does not contain a menu.


arundhati24:
@chiragw15 Would you please review this PR?

PR #1249, opened 2018-05-17
Remove menu from ChatSettingsFragment (Fixed #1243)

### LINKED ISSUE #1243: Remove menu from Settings Activity
**Actual Behaviour**

Currently, there is a menu in the Settings activity, which contains only one option i.e. About. This is not a correct UI flow because we reached the Settings activity from the Skills Activity menu, which has the options About as well as Settings. Now, we allow the user to reach the About Activity from the Settings activity. Also, once we reach About activity from Settings, the back button should direct us to Settings activity, which is not the case as of now.
Consider a situation, when we might have more number of options in our menu. Then, also it would not be correct to have a sub-menu like this in each activity.

**Expected Behaviour**

The Settings activity doesn't need a menu. If the user wishes to open About Activity, s/he should press back button to return to the SkillsActivity from where s/he can choose the About option from the menu.

**Screenshots of the issue**

about_in_settings

Yes

arundhati24:
Fixes #1243 

Changes: Removed menu present in ChatSettingsFragment.


arundhati24:
@chiragw15 As I do not have write access yet, would you please trigger the CircleCI build once again? (Some of the PRs have been passing CI tests just by triggering the build process again (by adding spaces here and there etc.) )

PR #1253, opened 2018-05-18
Disabled Tests

### LINKED ISSUE #1181: Circle CI Build Failing
**Actual Behaviour**

Circle CI build is failing most of the time
```
java.lang.RuntimeException: Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus. Root:
```

**Expected Behaviour**

Seems like the root view isn't getting focus as some other screen might be obstructing it. In short, the UI tests are failing and need to be fixed.

No. Comment if you want to work.

batbrain7:
Fixes #1181 

Changes: Disabled the UI tests

@arundhati24 @chiragw15 please review


batbrain7:
reverting the commit @iamareebjamal

PR #1255, opened 2018-05-18
Replaced hard-coded string by string resource value (Fixed #1148)

### LINKED ISSUE #1148: Replace hard-coded strings by string resource values
**Actual Behaviour**

Currently, there are a few layout resource files that contain hard-coded strings.

**Expected Behaviour**

The strings should not be hard-coded into the layout resource files. They should be referenced from string resource file strings.xml .

Yes

arundhati24:
Fixes #1148 

Changes: Replaced hard-coded string by string resource value.


iamareebjamal:
@arundhati24 I'm sure there are many more instance of hardcoded strings in the project. You may handle this later as this is low priority for now

arundhati24:
@iamareebjamal sure.

PR #1256, opened 2018-05-18
Removed the connectedCheck

### LINKED ISSUE #1181: Circle CI Build Failing

batbrain7:
Fixes #1181

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,11 +19,9 @@ test:
             parallel: true
         
         # Wait for it to have booted
-        - circle-android wait-for-boot
-        
+
         # Tun tests  against the emulator.
-        - ./gradlew connectedCheck
-        
+
         # Run Lint
         - ./gradlew lint
         
--- a/circle.yml
+++ b/circle.yml
@@ -22,7 +22,7 @@ jobs:
           command: ./gradlew lint
       - run:
           name: Test Coverage
-          command: ./gradlew testDebugUnitTestCoverage
+          command: ./gradlew build
       - store_artifacts:
           path: app/build/reports
           destination: reports
--- a/circle.yml
+++ b/circle.yml
@@ -23,6 +23,11 @@ jobs:
       - run:
           name: Run Tests
           command: ./gradlew build
+      - run:
+          command:
+            bash exec/prep-key.sh
+            chmod +x exec/apk-gen.sh
+            ./exec/apk-gen.sh
       - store_artifacts:
           path: app/build/reports
           destination: reports
@@ -31,7 +36,3 @@ jobs:
           destination: outputs
       - store_test_results:
           path: app/build/test-results
-      - run:
-          command: bash exec/prep-key.sh
-          command: chmod +x exec/apk-gen.sh
-          command: ./exec/apk-gen.sh
--- a/circle.yml
+++ b/circle.yml
@@ -1,38 +1,33 @@
-#Install android build tools, platforms
-#Supported versions here https://circleci.com/docs/android
-dependencies:
-    override:
-        - ( sleep 5 && while [ 1 ]; do sleep 1; echo y; done ) | android update sdk --no-ui --all --filter "tools,platform-tools,android-25,extra-google-m2repository,extra-android-m2repository,extra-android-support,extra-google-google_play_services"
-        - echo y | android update sdk --no-ui --all --filter "build-tools-26.0.2"
-        - ANDROID_HOME=/usr/local/android-sdk-linux ./gradlew dependencies
-
-#Pull any submodules
-#-PdisablePreDex is a must else gradle just dies due to memory limit
-#Replace
-test:
-    override:
-        - (./gradlew assemble -PdisablePreDex):
-            timeout: 360
-        - cp -r ${HOME}/${CIRCLE_PROJECT_REPONAME}/app/build/outputs/apk/ $CIRCLE_ARTIFACTS
-        - emulator -avd circleci-android22 -no-audio -no-window:
-            background: true
-            parallel: true
-        
-        # Wait for it to have booted
-
-        # Tun tests  against the emulator.
-
-        # Run Lint
-        - ./gradlew lint
-        
-        # Copy lint report
-        - mkdir $CIRCLE_TEST_REPORTS/Lint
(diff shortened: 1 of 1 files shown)


batbrain7:
@iamareebjamal please review

circle.yml
```diff
@@ -19,11 +19,9 @@ test:
             parallel: true
         
         # Wait for it to have booted
-        - circle-android wait-for-boot
-        
+
         # Tun tests  against the emulator.
-        - ./gradlew connectedCheck
-        
+
```
iamareebjamal: Remove comments, add `./gradlew build`

Remove
```
emulator -avd circleci-android22 -no-audio -no-window:
            background: true
parallel: true
```

iamareebjamal: Remove

```
- mkdir $CIRCLE_TEST_REPORTS/AndroidTests
        - cp -r app/build/reports/androidTests/connected/* $CIRCLE_TEST_REPORTS/AndroidTests
```

And replace this with unit tests report

iamareebjamal: Change this to correct build apk directory `cp -r ${HOME}/${CIRCLE_PROJECT_REPONAME}/app/build/outputs/apk/ $CIRCLE_ARTIFACTS`

iamareebjamal:
Also add caching, see Orga App circle config for details

circle.yml
batbrain7: @iamareebjamal unit tests are not yet initialised so path are you referring to?

batbrain7: which path*

iamareebjamal: Unit tests are ran when build command is executed

batbrain7: so the reports path will ci path starting with usr/ubuntu/  ?

iamareebjamal: No, just the relative path will change. Please take a look at Orga App config

batbrain7:
@chiragw15 the config file here is a little different than from the orga app's file
Please assist here

iamareebjamal:
@batbrain7 Just look at the test report path. Also look at the circle CI config format. Orga App is using v2 of that. You can use the same

batbrain7:
@iamareebjamal updating the config .yml file completely then

circle.yml
```diff
@@ -1,40 +1,33 @@
...
-        - mkdir $CIRCLE_TEST_REPORTS/Lint
-        - cp app/build/reports/lint-results.* $CIRCLE_TEST_REPORTS/Lint
-
-        # Copy tests report
-        - mkdir $CIRCLE_TEST_REPORTS/AndroidTests
-        - cp -r app/build/reports/androidTests/connected/* $CIRCLE_TEST_REPORTS/AndroidTests
-
-        - bash exec/prep-key.sh
-        - chmod +x exec/apk-gen.sh
-        - ./exec/apk-gen.sh
```
iamareebjamal: This should be added to upload the apk

batbrain7:
@iamareebjamal please review and merge

iamareebjamal:
Strange that unit tests did not run on build. Try locally and check what's happening

iamareebjamal:
The report for JUnit tests should be generated on build. If it is not being generated, then explicitly run `./gradlew test`

batbrain7:
@iamareebjamal build successfully passed on ./gradlew test command, should i add this command ?

iamareebjamal:
It should not just pass, there should be generated test report in the build folder as well

batbrain7:
@iamareebjamal  in the app there is only an androidTest package, I found it weird not finding a test package

iamareebjamal:
That means even unit tests are in androidTest package, please move them locally and try

iamareebjamal:
It seems like there are no unit tests at all. Just fix the line and change the requested for now. Thanks

batbrain7:
Yes @iamareebjamal there are no unit tests at all

circle.yml
```diff
@@ -1,40 +1,33 @@
...
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+      - run:
+          name: Download Dependencies
+          command: ./gradlew androidDependencies
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+      - run:
+          name: Run Tests
```
iamareebjamal: Change to `lint check`

circle.yml
```diff
@@ -1,40 +1,33 @@
...
+          command: ./gradlew lint
+      - run:
+          name: Test Coverage
```
iamareebjamal: Change to Run tests

circle.yml
iamareebjamal: This is not resolved

circle.yml
```diff
@@ -1,40 +1,36 @@
...
+          path: app/build/reports
+          destination: reports
+      - store_artifacts:
+          path: app/build/outputs
+          destination: outputs
+      - store_test_results:
+          path: app/build/test-results
+      - bash exec/prep-key.sh
+      - chmod +x exec/apk-gen.sh
+      - ./exec/apk-gen.sh
```
iamareebjamal: This is not how it works. You have to add it in run block in command section like given above for tests and lint

circle.yml
batbrain7: in what format this be added? I tried adding these at the end but ci build showed 
this error : orb not found: bash exec

circle.yml
batbrain7: Doing the changes

batbrain7:
@iamareebjamal CI fail shows permission denied error could adding su or sudo fix it?

batbrain7:
@iamareebjamal done the changes

batbrain7:
@iamareebjamal please merge this, now we can proceed towards the Unit tests

PR #1272, opened 2018-05-22
Apk upload modifications 

### LINKED ISSUE #1180: APKs in apk branch not updating
**Actual Behaviour**

Ever since PR #1033 merge, the APKs in apk branch are not updating. 
ref All Circle CI merge builds after [this](https://circleci.com/gh/fossasia/susi_android/2262) build are failing.

**Expected Behaviour**

Currently, do not know the reason. But expected behaviour should obviously be that the APKs should update. 

**LogCat for the issue**
```
./exec/apk-gen.sh

apk    build.gradle  exec		gradlew      nohup.out
app    circle.yml    gradle		gradlew.bat  README.md
build  docs	     gradle.properties	LICENSE      settings.gradle
cp: cannot stat ‘/home/ubuntu/susi_android/app/build/outputs/apk/app-debug.apk’: No such file or directory

./exec/apk-gen.sh returned exit code 1
```

yes

batbrain7:
Fixes #1180 

Changes: added three separate run tasks in circleci config file

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/code
     docker:
-      - image: circleci/android:api-26-alpha
+      - image: circleci/android:api-27-alpha
     environment:
       JVM_OPTS: -Xmx3200m
     steps:
--- a/circle.yml
+++ b/circle.yml
@@ -26,9 +26,9 @@ jobs:
       - run:
           command:
             bash exec/prep-key.sh
-            chmod +x exec/apk-gen.sh
       - run:
           command:
+            chmod +x exec/apk-gen.sh
             ./exec/apk-gen.sh
       - store_artifacts:
           path: app/build/reports
--- a/circle.yml
+++ b/circle.yml
@@ -25,11 +25,10 @@ jobs:
           command: ./gradlew build
       - run:
           command:
-            bash exec/apk-gen.sh
+            bash exec/prep-key.sh
       - run:
           command:
-            chmod +x exec/apk-gen.sh
-            ./exec/apk-gen.sh
+            bash exec/apk-gen.sh
       - store_artifacts:
           path: app/build/reports
           destination: reports
--- a/circle.yml
+++ b/circle.yml
@@ -26,8 +26,6 @@ jobs:
       - run:
           command:
             bash exec/prep-key.sh
-      - run:
-          command:
             chmod +x exec/apk-gen.sh
       - run:
           command:
--- a/circle.yml
+++ b/circle.yml
@@ -26,7 +26,11 @@ jobs:
       - run:
           command:
             bash exec/prep-key.sh
+      - run:
+          command:
             chmod +x exec/apk-gen.sh
+      - run:
+          command:
             ./exec/apk-gen.sh
       - store_artifacts:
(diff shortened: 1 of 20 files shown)


batbrain7:
screen shot 2018-05-22 at 11 23 11 am

@iamareebjamal this is shown in the circle ci config?

iamareebjamal:
This is not related to build scripts. This is circle's artifact upload system

circle.yml
```diff
@@ -26,7 +26,11 @@ jobs:
       - run:
           command:
             bash exec/prep-key.sh
+      - run:
+          command:
             chmod +x exec/apk-gen.sh
+      - run:
+          command:
```
iamareebjamal: Merge these two into one like `bash exec/apk-gen.sh`

circle.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       - run:
           command:
             bash exec/prep-key.sh
+      - run:
+          command:
             chmod +x exec/apk-gen.sh
             ./exec/apk-gen.sh
```
iamareebjamal: Not this. `bash exec/apk-gen.sh`

batbrain7: @iamareebjamal I combined it in the second commit but the CI showed an error of permission denied

circle.yml
```diff
@@ -25,7 +25,9 @@ jobs:
           command: ./gradlew build
       - run:
           command:
-            bash exec/prep-key.sh
+            bash exec/apk-gen.sh
+      - run:
+          command:
             chmod +x exec/apk-gen.sh
             ./exec/apk-gen.sh
```
iamareebjamal: Still same as previous commit

Should be `bash exec/apk-gen.sh`

circle.yml
```diff
@@ -25,7 +25,9 @@ jobs:
           command: ./gradlew build
       - run:
           command:
-            bash exec/prep-key.sh
+            bash exec/apk-gen.sh
```
iamareebjamal: This should be `bash exec/prep-key.sh`

circle.yml
batbrain7: removing statements 
chmod +x exec/apk-gen.sh
./exec/apk-gen.sh

and adding bash exec/apk-gen.sh ?

iamareebjamal: Yes

batbrain7:
@iamareebjamal can you please tell what is the difference b/w the statements ?

iamareebjamal:
bash is interpreter just like python. If you want to run python script, you write `python script.py` it runs even if there is no executable permission on the file because you are explicitly invoking python which itself is executable

Whereas, you can also run the python script as `./script.py` only you need shebang telling `#!/bin/python` on top so that the shell knows which interpreter to send the script to. Also, you need to have executable permission on the file. This is very restricted with shebang and permission, that's why it's just better to use `bash script.sh` or `python script.py`

iamareebjamal:
One more change is required. The apk-gen script does not have the correct version of build tools in it. Please fix it

batbrain7:
@iamareebjamal where to find the apk-gen script?

iamareebjamal:
You have written its path in the run command in build script

PR #1289, opened 2018-05-24
Replaced Customer server with custom server

### LINKED ISSUE #1192: Replace "Customer Server" with "Custom Server"
**Actual Behaviour**

Custom Server is wrongly written as Customer Server which can be misleading.

**Expected Behaviour**

It should be "Custom Server"

**Screenshots of the issue**

image

Yes

batbrain7:
Fixes #1192 

Changes: changed customer server to custom server

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/forgotpassword/ForgotPasswordActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/forgotpassword/ForgotPasswordActivity.kt
@@ -84,8 +84,8 @@ class ForgotPasswordActivity : AppCompatActivity (), IForgotPasswordView {
     }
 
     fun showUrl() {
-        customer_server.setOnClickListener {
-            input_url.visibility = if(customer_server.isChecked) View.VISIBLE else View.GONE
+        custom_server.setOnClickListener {
+            input_url.visibility = if(custom_server.isChecked) View.VISIBLE else View.GONE
         }
     }
 
@@ -98,7 +98,7 @@ class ForgotPasswordActivity : AppCompatActivity (), IForgotPasswordView {
     fun requestPassword() {
         reset_button.setOnClickListener {
             val email = forgot_email.editText?.text.toString()
-            val isPersonalServerChecked = customer_server.isChecked
+            val isPersonalServerChecked = custom_server.isChecked
             val url = input_url.editText?.text.toString()
             forgot_email.error = null
             input_url.error = null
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/fossasia/susi/ai/forgotpassword/ForgotPasswordActivity.kt
```diff
@@ -84,8 +84,8 @@ class ForgotPasswordActivity : AppCompatActivity (), IForgotPasswordView {
     }
 
     fun showUrl() {
-        customer_server.setOnClickListener {
-            input_url.visibility = if(customer_server.isChecked) View.VISIBLE else View.GONE
+        custom_server.setOnClickListener {
```
iamareebjamal: Change to camel case

app/src/main/res/values/strings.xml
```diff
@@ -188,7 +188,7 @@
     <!-- Settings Activity-->
     <string name="Continue">Continue</string>
 
-    <string name="email_id_input">Enter your e-mail ID</string>
+    <string name="email_id_input">Enter your e-mail address</string>
```
iamareebjamal: Irrelevant to PR

app/src/main/res/values-hi/strings.xml
```diff
@@ -185,7 +185,7 @@
     <string name="slide_3_desc">ब्राउज़ करें और अपने पसंदीदा एसयूएसआई एआई कौशल की कोशिश करें</string>
     <string name="slide_4_title">चैट सेटिंग्स</string>
     <string name="slide_4_desc">बेहतर अनुभव के लिए अपनी चैट सेटिंग्स को निजीकृत करें</string>
-    <string name="customer_server">ग्राहक सर्वर</string>
```
iamareebjamal: Check other strings as well

app/src/main/java/org/fossasia/susi/ai/forgotpassword/ForgotPasswordActivity.kt
chiragw15: It is not a variable name but a resource id which is directly used as a variable using kotlinx. Should it be changed to camelCase?

iamareebjamal: Yes, it's kotlin and java convention to use camel case

batbrain7: @iamareebjamal should this change be made in a separate PR because we have to make these changes in every activity?

iamareebjamal: OK

PR #1296, opened 2018-05-24
Increment version code (Fixed #1295)

### LINKED ISSUE #1295: Update version code and name
Update the version code and version name for the app for version 1.0.6 released as alpha.

arundhati24:
Fixes #1295 

Changes: Increment version code for app for version 1.0.6 released as alpha.


app/build.gradle
```diff
@@ -32,7 +32,7 @@ android {
         applicationId "ai.susi"
         minSdkVersion rootConfiguration.minSdkVersion
         targetSdkVersion rootConfiguration.targetSdkVersion
-        versionCode 7
+        versionCode 8
         versionName "1.0.6"
```
iamareebjamal: Change

arundhati24: @iamareebjamal Do I need to change the versionName? The version is 1.0.6 that is why I did not change it.

iamareebjamal: Yes both versionCode and versionName need to change with each dev cycle

arundhati24: @iamareebjamal The latest version on play store is 1.0.6. The version name in the file was already 1.0.6, therefore I did not change it. If it needs to change, it would be great if you could please let me know what should i name it as.

screen shot 2018-05-25 at 10 24 33 pm

iamareebjamal: versions are changed **before** the app is uploaded to Play Store. If you check the commit history, I already changed 5 -6 and 1.0.6, so naturally for next development cycle, this needs to change to 1.0.7

arundhati24: @iamareebjamal Thanks for the clarification. :)
I will do that asap.

arundhati24: @iamareebjamal This has been fixed already. Would you please have a look?

Issue #1311, opened 2018-05-27
Refactor code


arundhati24:
Here are a few things that are highlighted on running lint check and need to be fixed :

- [ ] Define non-translatable resources in base values folder
- [ ] Incomplete translations. Add translations for other specified languages as well.
- [ ] Fix invalid string formats that have been used
- [ ] Use app:srcCompat when using Vector Drawables
- [ ] Fix lint accessibility warnings
- [ ] Unregistered classes in Android Manifest
- [ ] Layout inflation without a parent
- [ ] Fix typos
- [x] Fix warnings in xml (e.g. unused namespace declaration, general Internationalisation issues like using left/right instead of start/end, Padding and Margin symmetry, Use sp instead of dp for text sizes, hardcoded strings)
- [ ] Useless parent layouts present (transfer background to other view)
- [ ] Handle reference leaks
- [ ] Inefficient layout weights
- [ ] Unused resources present
- [ ] Fix lint security issues
- [ ] Fix lint usability issues
- [x] Fix redundant local variables and constructs and remove unused imports
- [ ] Use of redundant or deprecated syntax or symbols
- [x] Fix probable bugs (e.g. @NotNull/ @Nullable problems)
- [ ] Fix style issues

**Would you like to work on the issue?**

Yes

PR #1318, opened 2018-05-28
Fixed probable bugs (Fixed #1317)

### LINKED ISSUE #1317: Fix probable bugs
[Parent issue #1311 ]

**Actual Behaviour**

Currently, there are a number of warnings on running lint check that might produce exceptions :
- Constant conditions and expressions that are always true or false
- Problems related to @Nullable and @NotNull annotations usage in constant conditions & exceptions
- Unused assignments

**Expected Behaviour**

These conditions might result in exceptions like Null Pointer Exception. Hence, there is a need to use proper annotations and a need to handle conditions and expressions properly to avoid probable bugs.

Yes

arundhati24:
Fixes #1317 [Parent #1311 ]

Changes: Added appropriate annotations and checked for nullability conditions, wherever required.

### DIFF
--- a/app/src/main/java/ai/kitt/snowboy/AppResCopy.java
+++ b/app/src/main/java/ai/kitt/snowboy/AppResCopy.java
@@ -45,7 +45,7 @@ public class AppResCopy {
                 InputStream is = context.getAssets().open(assetsSrcDir);
                 FileOutputStream fos = new FileOutputStream(outFile);
                 byte[] buffer = new byte[1024];
-                int byteCount = 0;
+                int byteCount;
                 while ((byteCount = is.read(buffer)) != -1) {
                     fos.write(buffer, 0, byteCount);
                 }
--- a/app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
@@ -92,7 +92,7 @@ public class ChatFeedRecyclerAdapter extends RealmRecyclerViewAdapter<ChatMessag
         lastMsgCount = getItemCount();
         RealmChangeListener<RealmResults> listener = new RealmChangeListener<RealmResults>() {
             @Override
-            public void onChange(RealmResults elements) {
+            public void onChange(@NonNull RealmResults elements) {
                 //only scroll if new is added.
                 if (lastMsgCount < getItemCount()) {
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/ChatFeedRecyclerAdapter.java
```diff
@@ -221,6 +222,7 @@ public void onDetachedFromRecyclerView(RecyclerView recyclerView) {
     public int getItemViewType(int position) {
         ChatMessage item = getItem(position);
 
+        assert item != null;
```
iamareebjamal: Never use assert in production code

iamareebjamal: We don't need to throw exceptions. We need to check if they are null and handle it gracefully

If they should not be null and they are, then well any other exception is not much better than NullPointerException

Mostly, exceptions are good in service layer, never in view layer

arundhati24:
@iamareebjamal Made the changes.

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.java
```diff
@@ -244,12 +258,16 @@ public void onFailure(Call<SkillRatingResponse> call, Throwable t) {
...
-                        setRating(false, true);
+                        if(thumbsUp != null) {
+                            thumbsUp.setImageResource(R.drawable.thumbs_up_outline);
+                            setRating(false, true);
+                        }
                         break;
                     case Constant.NEGATIVE:
-                        thumbsDown.setImageResource(R.drawable.thumbs_down_outline);
-                        setRating(false, false);
+                        if(thumbsDown != null) {
```
iamareebjamal: Space after if is missing everywhere

app/src/main/java/ai/kitt/snowboy/AppResCopy.java
```diff
@@ -45,7 +45,7 @@ private static void copyFilesFromAssets(Context context, String assetsSrcDir, St
                 InputStream is = context.getAssets().open(assetsSrcDir);
                 FileOutputStream fos = new FileOutputStream(outFile);
                 byte[] buffer = new byte[1024];
-                int byteCount = 0;
+                int byteCount ;
```
iamareebjamal: Space after ; should not be there

PR #1341, opened 2018-06-02
Fix  #1339 and refactored code

### LINKED ISSUE #1339: Seach bar from Skills Activity appear in Settings
**Actual Behaviour**

Searchbar and keyboard appears in the Settings screen, in a particular sequence.

**Expected Behaviour**

Search bar and keyboard should be closed

**Steps to reproduce it**

1. Tap on the Search icon in Skills Activity ,then the keyboard gets opened in the Skills Activity.
2. Tap on Settings, without doing anything else.
3. It takes us to the Settings page, but without closing the Keyboard and Search bar.

**Screenshots of the issue**

bug

Yes

EDIT: Happens for SkillDetails too

agh372:
Fixes #1339

Changes:
MVP pattern was not being followed in Skill Activity, so refactored that too, and added a callback so that SkillDetailsFragment loading gets called in SkillActivity class

@iamareebjamal @batbrain7 please review

### DIFF
--- a/app/src/androidTest/java/org/fossasia/susi/ai/skills/SkillsActivityTest.kt
+++ b/app/src/androidTest/java/org/fossasia/susi/ai/skills/SkillsActivityTest.kt
@@ -9,6 +9,7 @@ import android.support.test.rule.ActivityTestRule
 import android.support.test.runner.AndroidJUnit4
 import android.view.WindowManager
 import org.fossasia.susi.ai.R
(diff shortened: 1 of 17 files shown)


iamareebjamal:
This PR is handling too much at once. What's the advantage of creating a separate package just for activity

agh372:
I have added presenter and view interfaces too for that activity, so I thought of creating a separate package for it, then what can I do for that?

iamareebjamal:
Revert to the same package and split the PR to handle only single issue

PR #1346, opened 2018-06-03
Use same styles to name strings in Settings Activity (Fixed #1345)

### LINKED ISSUE #1345: Use uniform cases for strings in the Settings Activity
**Actual Behaviour**

Currently, there are different cases used in the settings strings like "Speech Settings" written in title case while "SUSI Voice settings" does not follow title case.

**Expected Behaviour**

Use same conventions for all the string. (First letter of first word in upper case, rest all letters in lowercase).

Yes

arundhati24:
Fixes #1345 

Changes: All the strings in settings activity follow the same naming style. (First letter of first word will be uppercase while all others will be lowercase)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -56,11 +56,11 @@
     <string name="send_message">Sending the message</string>
     <string name="send_msg_hint">Ask SUSI something…</string>
 
-    <string name="server_pref">Select Server</string>
+    <string name="server_pref">Select server</string>
     <string name="server_select_summary">Select backend server for the app.</string>
-    <string name="server_settings_title">Server Settings</string>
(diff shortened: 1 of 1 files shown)


arundhati24:
@iamareebjamal I tried to follow these guidelines. Also, there was no uniform style. Few of the labels had only the first word capitalized while others had all words capitalized. 
It would be helpful if you could please reflect upon what should be the final style to be used so that the non-uniformity could be removed.

screen shot 2018-06-03 at 10 09 14 pm


screen shot 2018-06-03 at 10 09 25 pm

iamareebjamal:
You are correct but `Chat settings` or `Chat Settings`. I think second should be preferred, but yes guidelines say the opposite

So, let's go with the guidelines

Issue #1362, opened 2018-06-06
Hide the rating bar when the user is not logged in


arundhati24:
**Actual Behaviour**

Currently, the rating bar is displayed even when the user is not logged in. It does not make sense to display the rating bar in this case because without a user identity (access token) a skill cannot be rated.

**Expected Behaviour**

Hide the rating bar when the user chooses to use the app anonymously and show the rating bar when the user is logged in.

**Would you like to work on the issue?**

Yes

PR #1364, opened 2018-06-06
Change the data type of key values in stars object (Fixed #1363)

### LINKED ISSUE #1363: Change the data type of stars as per the recent change in the API
**Actual Behaviour**

The data type of stars object values were String earlier, now they have been changed to Int and Float as per their requirement and use cases. Hence, the app does not show the skill rating properly.

**Expected Behaviour**

Change the types of the stars object values as per the recent changes in the get SkillList API so as to show the proper skill rating statistics.

Yes

arundhati24:
Fixes #1363 

Changes: The data types of key values in the stars object of 'cms/getSkillList.json' API have changed from String to Int for all except Float for avg_star, hence the necessary updates have been made in the Android client.


arundhati24:
@iamareebjamal @batbrain7 Would you please review and merge this PR? It would be helpful as I need to work on #1362 after this where I also need to organise the code a bit so as to place them in the correct functions logically. Doing that without merging this PR would only lead to a set of unnecessary conflicts.

PR #1365, opened 2018-06-07
Hide rating bar when the user is not logged in (Fixed #1362)

### LINKED ISSUE #1362: Hide the rating bar when the user is not logged in

arundhati24:
Fixes #1362 

Changes: Hid the rating bar when the user is not logged in.

**Screenshot : When user is logged in**

screenshot_1528392083

screenshot_1528392095

**Screenshot : When user is not logged in**

screenshot_1528392649

screenshot_1528392138

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
@@ -26,6 +26,7 @@ import com.squareup.picasso.Picasso
 import kotlinx.android.synthetic.main.fragment_skill_details.*
 import org.fossasia.susi.ai.R
 import org.fossasia.susi.ai.chat.ChatActivity
+import org.fossasia.susi.ai.helper.PrefManager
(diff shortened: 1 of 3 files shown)


arundhati24:
@batbrain7 Would you please highlight the discrepancy?

batbrain7:
@arundhati24 in the first two screenshots, on the left side of the bars, there is 5* shown and below  it there is overlapping between the textviews

arundhati24:
@batbrain7 I agree with you and I am aware of this discrepancy. Actually this occurs because the labels on the bars are 0.0 and the bar width is zero. As a result, these labels overlap with the axis labels. This will not happen if other bars too have some width. I will be fixing this in another PR because I need to check the documentation for MPAndroidChart library and format the values accordingly. Hence, this is beyond the subject of this PR.

PR #1367, opened 2018-06-08
Added Hint in skill search

### LINKED ISSUE #1211: Search Functionality in Skill List Fragment implemented improperly
**Actual Behaviour**

Only first matching search is displayed rather than showing all the matching results. 

```
for (item2 in item.second.keys){
                if( query.toLowerCase() in item2){
                    skillGroups.scrollToPosition(pos)
                    return
                }
```
The above code snippet is currently in use for search a query. Firstly, complexity can be improved for efficient search and even the logic is not completely correct. This simply finds the given query in the name of the skill and scrolls to the first position it finds.

**Expected Behaviour**

To improve this, we can create a new list and find skills from the current list matching the given query and then we can transfer those skills to a new list and display that list instead of just showing the top element which matches the query.

Yes

batbrain7:
Fixes #1211 

Changes: Added the hint and removed the underline

Screenshots for the change: 
screen shot 2018-06-08 at 8 08 41 am

### DIFF
--- a/app/src/main/res/layout/search_bar.xml
+++ b/app/src/main/res/layout/search_bar.xml
@@ -8,11 +8,11 @@
         android:id="@+id/edtSearch"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_marginTop="10dp"
+        android:layout_marginTop="@dimen/search_skill_margin"
         android:background="@null"
         android:hint="@string/search_skill"
         android:imeOptions="actionSearch"
         android:inputType="text"
-        android:padding="7dp" />
+        android:padding="@dimen/search_skill_padding" />
 
 </LinearLayout>
--- a/app/src/main/res/layout/search_bar.xml
+++ b/app/src/main/res/layout/search_bar.xml
@@ -8,11 +8,11 @@
         android:id="@+id/edtSearch"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_marginTop="@dimen/search_skill_margin"
+        android:layout_marginTop="@dimen/margin_large"
         android:background="@null"
         android:hint="@string/search_skill"
         android:imeOptions="actionSearch"
         android:inputType="text"
-        android:padding="@dimen/search_skill_padding" />
+        android:padding="@dimen/margin_medium" />
 
 </LinearLayout>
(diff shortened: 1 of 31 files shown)


app/src/main/res/layout/search_bar.xml
```diff
@@ -8,7 +8,11 @@
         android:id="@+id/edtSearch"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
+        android:layout_marginTop="10dp"
```
arundhati24: Replace by dimen resource

app/src/main/res/layout/search_bar.xml
```diff
@@ -8,7 +8,11 @@
...
+        android:background="@null"
+        android:hint="@string/search_skill"
         android:imeOptions="actionSearch"
-        android:inputType="text" />
+        android:inputType="text"
+        android:padding="7dp" />
```
arundhati24: Replace by dimen resource
Also, if possible, please use standard padding values as per Material Design Guidelines.

app/src/main/res/values/dimens.xml
```diff
@@ -99,4 +99,7 @@
     <dimen name="scan_progress_margin_top">10dp</dimen>
     <dimen name="device_setup_margin_top">20dp</dimen>
 
+    <dimen name="search_skill_padding">7dp</dimen>
+    <dimen name="search_skill_margin">10dp</dimen>
```
arundhati24: Repeated margin value. Use margin_large already present in the file.

batbrain7: @arundhati24 it is the name that matters so that the code is easily readable

arundhati24: @batbrain7 There is another issue that I had opened that will be removing repeated dimens so as to use standard padding/margin values and naming conventions as per the Material Design Guidelines for a more consistent design. Hence, for now, we can replace this with margin_large that is already present in the file, so as to avoid the repetition of dimens. This will be more uniform once that issue is fixed.

app/src/main/res/values/dimens.xml
```diff
@@ -99,4 +99,7 @@
     <dimen name="scan_progress_margin_top">10dp</dimen>
     <dimen name="device_setup_margin_top">20dp</dimen>
 
+    <dimen name="search_skill_padding">7dp</dimen>
```
arundhati24: Is it possible to use 8dp instead of 7dp?

batbrain7: I tried 8dp but it looked very wide in small phones, so I think 7dp is better

iamareebjamal: You should be using standard dimensions under material guidelines.

batbrain7:
@iamareebjamal @arundhati24 made the changes

Issue #1366, opened 2018-06-07
Fix discrepancies in skill rating chart


arundhati24:
**Actual Behaviour**

Currently, the bar labels overlap the axis labels if the bar width is 0.0 or less than the width required to fit in the bar label.

**Expected Behaviour**

The bar labels should not overlap the axis labels.

**Would you like to work on the issue?**

Yes

arundhati24:
Solved in #1380

PR #1380, opened 2018-06-11
Improve UI for skill rating section (Fixed #1379)

### LINKED ISSUE #1379: Improve UI of skill rating section in skill details screen
**Actual Behaviour**

Currently, the UI had a number of discrepancies and it was not similar to that used in other platforms.

**Expected Behaviour**

The UI has been polished to remove existing discrepancies and to match the UI of the skills CMS.

Yes

arundhatigupta:
Fixes #1379 
(Also addresses #1366 )

Changes: Improved the UI to remove existing discrepancies and to make it look similar to that used on Skills CMS and iOS.

Screenshots for the change: 

screenshot_1528713298


arundhati24:
@batbrain7 It is the material star icon that I have used as a label for the graph.

batbrain7:
@arundhati24 make a vector drawable instead of png file

arundhati24:
@batbrain7 For now, if that doesn't affect the performance much, would it be possible to please keep it? It's working fine. I would update it soon in another PR. I first need to update the user rating as soon as the user rates a skill. And, that has been throwing some errors. Hence, that needs to be fixed first.

iamareebjamal:
@arundhati24 Thanks for the contribution. One big problem with committing binary files is that it increases the size of the repository monotonically. It can never be decreased from that point since git can't store the changes as diff. So it's OK to use binary files in PR but convert to vectors before merging

arundhati24:
@iamareebjamal Thanks for the info. I would definitely take care of that from the next time. And we will also fix that soon so that performance or size issues do not arise.

PR #1393, opened 2018-06-13
Update ratings as soon as the user rates a skill (Fixed #1378)

### LINKED ISSUE #1378: Update ratings as soon as the user rates a skill
**Actual Behaviour**

Currently, if the user rates a skill, the data shown in the skill rating section doesn't update. It updates only when the user goes back to the chat activity and returns to the skills listing activity again and checks the skill rating after this. This is because a call to the getSkillList API is sent only when the skills listing activity is opened.

**Expected Behaviour**

The rating shown in the skill details section should be updated as soon as the user rates a skill.

Yes

arundhati24:
Fixes #1378 

A demo for can be seen here : [Link](https://drive.google.com/open?id=1UMlIIG5e-COCVYexxnTxGIGk9wH5_iv0)

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/data/SkillDetailsModel.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/data/SkillDetailsModel.kt
@@ -1,0 +1,56 @@
+package org.fossasia.susi.ai.data
+
+import org.fossasia.susi.ai.data.contract.ISkillDetailsModel
+import org.fossasia.susi.ai.rest.ClientBuilder
+import org.fossasia.susi.ai.rest.responses.susi.FiveStarSkillRatingResponse
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * Model of SkillDetails
+ * The M in MVP
+ * Stores all business logic
+ *
+ * @author arundhati24
+ *
+ */
+
+class SkillDetailsModel : ISkillDetailsModel {
+
+    private lateinit var updateRatingsResponseCall: Call<FiveStarSkillRatingResponse>
+
+    /**
+     * Posts a request the fiveStarRateSkill.json API
+     *
+     * @param map  A query map consisting of following key value pairs
+     *             model       Model of the skill (e.g. general)
+     *             group       Group of skill (e.g. Knowledge)
+     *             language    Language directory in which the skill resides (e.g. en)
+     *             skill       Skill Tag of object in which the skill data resides
+     *             rating      User rating to be sent to the server
+     *             accessToken Access token of logged in user
+     *             listener
+     *
+     */
+    override fun fiveStarRateSkill(map: Map<String, String>, listener: ISkillDetailsModel.OnUpdateRatingsFinishedListener) {
+
+        updateRatingsResponseCall = ClientBuilder().susiApi.fiveStarRateSkill(map)
+
+        updateRatingsResponseCall.enqueue(object : Callback<FiveStarSkillRatingResponse> {
+            override fun onResponse(call: Call<FiveStarSkillRatingResponse>, response: Response<FiveStarSkillRatingResponse>) {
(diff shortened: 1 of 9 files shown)


arundhati24:
@iamareebjamal Here are the code changes. Please have a look. Thanks :)

app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/FiveStarSkillRatingResponse.kt
```diff
@@ -1,12 +1,26 @@
...
+    @SerializedName("session")
+    @Expose
     val session: Session? = null
+
+    @SerializedName("accepted")
+    @Expose
     val accepted: Boolean = false
+
+    @SerializedName("message")
+    @Expose
```
iamareebjamal: I don't think you need SerializedName and Expose if property names are name, please confirm

iamareebjamal:
The logic seems too complicated for a simple use case. Why not use coroutines instead of AsyncTask

app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/FiveStarSkillRatingResponse.kt
arundhati24: @iamareebjamal That's right. Thanks. Actually these annotations were not there initially in this file. Now I understand why these were not there. I will remove these as all the variable names used here are same as the keys.

arundhati24:
Okay 👍 I will find out more about coroutines and make the necessary changes soon.

arundhati24:
@iamareebjamal I have used coroutines for the first time. I had referred to the following resources to understand Kotlin coroutines : [Kotlin/kotlinx.coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/ui/coroutines-guide-ui.md#the-problem-of-ui-freezes), [Kotlin/kotlin-coroutines](https://github.com/Kotlin/kotlin-coroutines)  
I had updated this PR but I am unable to get the app running as the CI builds are failing. I am unable to understand it. I might be missing on something small. It would be very helpful if you could please have a look at the code changes and suggest the necessary changes.

iamareebjamal:
`More than one file was found with OS independent path 'META-INF/main.kotlin_module'`

app/src/main/java/org/fossasia/susi/ai/data/contract/ISkillDetailsModel.kt
```diff
@@ -0,0 +1,22 @@
...
+ * @author arundhati24
+ */
+interface ISkillDetailsModel {
+
+    interface OnUpdateRatingsFinishedListener {
+        fun onError(throwable: Throwable)
+        fun onSkillDetailsModelSuccess(response: Response<FiveStarSkillRatingResponse>)
+    }
+
+    fun fiveStarRateSkill(model: String, group: String, language: String, skill: String, stars: String,
```
iamareebjamal: Too many params, make object instead

app/src/main/java/org/fossasia/susi/ai/data/SkillDetailsModel.kt
```diff
@@ -0,0 +1,56 @@
...
+     *             model       Model of the skill (e.g. general)
+     *             group       Group of skill (e.g. Knowledge)
+     *             language    Language directory in which the skill resides (e.g. en)
+     *             skill       Skill Tag of object in which the skill data resides
+     *             rating      User rating to be sent to the server
+     *             accessToken Access token of logged in user
+     *             listener
+     *
+     */
+    override fun fiveStarRateSkill(map: Map<String, String>, listener: ISkillDetailsModel.OnUpdateRatingsFinishedListener) {
```
iamareebjamal: Use object, maps aren't typesafe

arundhati24:
Closing this, as this has already been merged along with other PR. Will replace query map with object in a new PR.

PR #1395, opened 2018-06-13
Send device_type=Android in the chat.json query

### LINKED ISSUE #1406: Send device type in chat.json API
The type of device ie, "Android" is  required to get device wise usage analytics.

**Steps to reproduce it**

NA

**LogCat for the issue**

https://api.susi.ai/susi/chat.json?q=hello&language=en&access_token=PGbzC5zYQBQUuCrYrarqIuRe4lvjfn&latitude=30.1667&longitude=77.3

**Screenshots of the issue**

NA

Yes

AnupKumarPanwar:
Fixes #1406 
Send device type in chat.json API 

Changes: 
Modified getSusiMessage() and getTableSusiMessage() functions to send device_type="Android" in the query 

Screenshots for the change: 
NA

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
@@ -41,6 +41,7 @@ class ChatPresenter(chatActivity: ChatActivity) : IChatPresenter, IChatModel.OnR
     private var micCheck = false
     var latitude: Double = 0.0
     var longitude: Double = 0.0
+    var deviceType: String = "Android"
     private var source = Constant.IP
     private var isDetectionOn = false
     var check = false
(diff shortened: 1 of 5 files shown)


batbrain7:
@iamareebjamal please merge this

app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
```diff
@@ -41,6 +41,7 @@ class ChatPresenter(chatActivity: ChatActivity) : IChatPresenter, IChatModel.OnR
     private var micCheck = false
     var latitude: Double = 0.0
     var longitude: Double = 0.0
+    var deviceType: String = "Android"
```
iamareebjamal: Why not private val?

iamareebjamal: Also, type inference will get the type automatically, no need to specifically name it

AnupKumarPanwar: I've followed the existing code pattern, as used by the latitude and longitude since they relate to the same thing.

iamareebjamal: No problem, we should leave the code better than we found it, make it private val

PR #1411, opened 2018-06-17
Send country code and name in chat.json API (Fixed #1394)

### LINKED ISSUE #1394: Send country name and country code in chat.json API
**Actual Behaviour**

Only latitude and longitude are send as API parameters.

**Expected Behaviour**

Country code and country name are also required to get country wise usage analytics.

**Steps to reproduce it**

Check `\susi_android\app\src\main\java\org\fossasia\susi\ai\chat\ChatPresenter.kt`

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

I think @batbrain7 and @arundhati24  can solve it.

arundhati24:
Fixes #1394 

Changes: Now country code and name is also sent as a query in chat.json API.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/data/ChatModel.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/data/ChatModel.kt
@@ -51,10 +51,9 @@ class ChatModel : IChatModel {
         })
     }
 
-    override fun getSusiMessage(timezoneOffset: Int, longitude: Double, latitude: Double, source: String,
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/susi/ai/data/ChatModel.kt
```diff
@@ -72,9 +74,11 @@ class ChatModel : IChatModel {
     }
 
     override fun getTableSusiMessage(timezoneOffset: Int, longitude: Double, latitude: Double, source: String,
```
iamareebjamal: Too many params, convert to object

PR #1418, opened 2018-06-18
Modify custom server URL error message

### LINKED ISSUE #1416: The custom server URL error messages are misguiding and unclear
**Actual Behaviour**

Whenever we enter an invalid custom server URL, on clicking the login button unclear and vague messages are shown. mostly they are the error messages that are shown to the error response http protocol codes like 404, 503 etc.

**Expected Behaviour**

Clear messages should be shown so that the user knows what type of URL should be entered.

**Steps to reproduce it**

add any custom server url in the login page and hit login

Yes

batbrain7:
Fixes #1416 

Changes: Added the message for incorrect Custom server URL

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -196,8 +196,8 @@
 
     <string name="error">Error</string>
     <string name="error_password_matching">Passwords do not match</string>
-    <string name="error_custom_server">Please make sure that you enter the correct custom server URL and take into consideration protocol part in the URL,
-        for eg. "http" and "https" is not allowed</string>
+    <string name="error_custom_server">Please make sure that you enter the correct custom server URL and do not take into consideration protocol part in the URL,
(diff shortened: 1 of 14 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -196,7 +196,9 @@
 
     <string name="error">Error</string>
     <string name="error_password_matching">Passwords do not match</string>
-
+    <string name="error_custom_server">Please make sure that you enter the correct custom server URL and do not take into consideration protocol part in the URL,
```
iamareebjamal: Why not just remove protocol yourself if found

batbrain7:
@iamareebjamal actually this is the wrong message that I put, the protocol part is already handled by the app and so this message only for invalid custom URL

batbrain7:
@iamareebjamal please review

PR #1433, opened 2018-06-22
Fix app crash in skill details screen (Fixed #1432)

### LINKED ISSUE #1432: App crashes due on opening skill details screen when user is not logged in
**Actual Behaviour**

Currently, the app crashes because the API is being called without checking for the log in status of the user.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

* Skip login
* Go to the skill listing screen
* Tap any skill card
App crashes.

Yes

arundhati24:
Fixes #1432 
Changes: Fixed app crash.


iamareebjamal:
@arundhati24 Please be more specific in your future commits

PR #1452, opened 2018-06-27
Show skills in the language chosen by the user (Fixed #1142)

### LINKED ISSUE #1142: Change of language in settings does not reflect
**Actual Behaviour**

Currently, changing the language from default to any other language does not reflect in the app i.e. if the language id set to German (say), there is no change in the language in the app. I remains set to default.

**Expected Behaviour**

If the user sets the language to any other language in the settings, the app should display data in that language.

**Screenshots of the issue**

language_change

Yes

arundhati24:
Fixes #1142 

Changes: Now the skills are shown in the skill listing activity in the language chosen by the user.

**Example : When the user chooses Hindi from Settings**

screenshot_1530078811

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
+++ b/app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
@@ -37,7 +37,7 @@ public class Constant {
     public static final String SELECT_SERVER = "Server_Select";
     public static final String LANG_SELECT = "Lang_Select";
     public static final String LANGUAGE = "prefLanguage";
-    public static final String DEFAULT = "default";
+    public static final String DEFAULT = "en";
     public static final String RATE = "rate";
     public static final String DEVICE = "device";
     public static final String DEVICE_SETUP = "setup";
--- a/app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
@@ -1,6 +1,7 @@
 package org.fossasia.susi.ai.skills.settings
 
 import android.Manifest
+import android.app.Activity
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.net.Uri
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
```diff
@@ -218,7 +222,7 @@ class ChatSettingsFragment : PreferenceFragmentCompat(), ISettingsView {
     }
 
     private fun setLanguage() {
-        val index = querylanguage.findIndexOfValue(PrefManager.getString(Constant.LANGUAGE, Constant.DEFAULT))
+        val index = querylanguage.findIndexOfValue(PrefManager.getString(Constant.LANGUAGE, "en"))
```
iamareebjamal: Rather change the constant than changing it everywhere

app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
```diff
@@ -37,7 +37,7 @@
     public static final String SELECT_SERVER = "Server_Select";
     public static final String LANG_SELECT = "Lang_Select";
     public static final String LANGUAGE = "prefLanguage";
-    public static final String DEFAULT = "default";
+    public static final String DEFAULT = "en";
```
batbrain7: @codedsun this is the reason for the crash on opening settings

app/src/main/res/values/array.xml
```diff
@@ -1,20 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string-array name="languagentries">
-        <item>Default</item>
```
batbrain7: this right here

app/src/main/java/org/fossasia/susi/ai/helper/Constant.java
arundhati24: @batbrain7 @codedsun Thanks for pointing out this issue. I see this wasn't reviewed well. I request you to please review the PRs properly before they are merged. Also, there wasn't any crash while this was fixed. In fact, I received a crash when I kept default rather than en. I was asked to change the constant rather than making changes everywhere. And, after doing that too there was no crash. So, it would be great if you review the code well and understand that without proper testing, there are always higher chances of getting bugs and crashes, and please don't approve any such PR where you have a doubt.

app/src/main/res/values/array.xml
arundhati24: Thanks, are you going to fix this or should I take this up and fix it quickly?

batbrain7: changing this won't solve the issue, please do proper research before suggesting any changes.
Thanks.

PR #1483, opened 2018-07-03
fix: Remove unsupported action types

### LINKED ISSUE #1482: Remove unsupported response types from the app
**Actual Behaviour**

The RSS action and some other action types have been discontinued from the server and hence is added redundantly in the app and is causing issues while refactoring the code.

**Expected Behaviour**

Remove the redundant action as it is not supported by the server anymore.

Yes

batbrain7:
Fixes #1482
 
Changes:  Many action types are not supported by the server, so it is removed.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
@@ -65,13 +65,6 @@ class ParseSusiResponseHelper {
                 null
             }
 
-            Constant.RSS -> datumList = try {
-                susiResponse.answers[0].data
-            } catch (e: Exception) {
-                Timber.e(e)
-                null
-            }
-
             Constant.WEBSEARCH -> webSearch = try {
                 susiResponse.answers[0].actions[1].query
             } catch (e: Exception) {
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
@@ -58,20 +58,6 @@ class ParseSusiResponseHelper {
                 null
             }
 
-            Constant.PIECHART -> datumList = try {
-                susiResponse.answers[0].data
-            } catch (e: Exception) {
-                Timber.e(e)
-                null
-            }
-
-            Constant.WEBSEARCH -> webSearch = try {
-                susiResponse.answers[0].actions[1].query
-            } catch (e: Exception) {
-                Timber.e(e)
-                ""
-            }
-
             Constant.STOP -> try {
                 stop = susiResponse.answers[0].actions[1].type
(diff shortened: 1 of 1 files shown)


batbrain7:
@codedsun Please review !

app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
```diff
@@ -65,13 +65,6 @@ class ParseSusiResponseHelper {
                 null
             }
 
-            Constant.RSS -> datumList = try {
-                susiResponse.answers[0].data
-            } catch (e: Exception) {
-                Timber.e(e)
-                null
-            }
-
```
iamareebjamal: Is it just this?

batbrain7: @iamareebjamal the viewholders and adapters related to this action are also used by other actions so only this part needs to be removed

batbrain7:
@arundhati please review this

batbrain7:
@arundhati24 I checked with @hkedia321 and he told me that RSS is not supported for months now also web search is also not supported. if the code should be kept in the app for future then we should only remove the parts where it reflects some response to the app.

arundhati24:
@batbrain7 Could you please link me to the discussion you had with @hkedia321 regarding this?

batbrain7:
@arundhati24 you may ask him yourself, as all discussions do not take place on the issue tracker

hkedia321:
@batbrain7 Though some advanced action types have stopped coming from the server (see [susi_server/issues/593](https://github.com/fossasia/susi_server/issues/593) and [susi_server/issues/726](https://github.com/fossasia/susi_server/issues/726)), I don't think so we should remove the support from the client sides for these. The advanced response types may come back. Also we have support for these in chat.susi.ai. @arundhati24 What do you think?

batbrain7:
Thanks @hkedia321 for the input so I think the support should not be removed, but should not be not detected as of now in the client and can be detected later on when the server supports these

batbrain7:
@iamareebjamal @arundhati24 please review

arundhati24:
@hkedia321 @akshatnitd I agree with both of you and I think these should not be removed from the client for now.

arundhati24:
@batbrain7 If the discussion has not taken place on issue tracker, it would have taken place on the Gitter channel. As I couldn't find any such discussion on the project channels, I asked you to link me to the discussion. How should I know when you had asked and what you had asked without reading the conversation. If you had asked it on a personal chat, that's a different case.

batbrain7:
@akshatnitd @arundhati24 I have not deleted the implementation but instead aligned it with the web client if you closely look at the PR, now whenever we ask any response of the type rss or web search it shows an error message just as in case of the web client , and since the web client has more features than the android client I think the goal should be to align them together,
Also if there is no other reason then please review this PR and merge it.

batbrain7:
@iamareebjamal please merge

PR #1487, opened 2018-07-03
Removed extra call to setRating() (Fixed #1486)

### LINKED ISSUE #1486: Extra call to setRating()
**Actual Behaviour**

There is an extra call to setRating().

**Expected Behaviour**

Remove this extra call as this is not needed and might lead to unnecessary issues.

Yes

arundhati24:
Fixes #1486 

Changes: Removed extra call to setRating()


arundhati24:

arundhati24:
@batbrain7 setRating() is already being called after the five star rating is successfully sent and an updated non-null object is received in the response. I have checked it. But, if you see any possibilities of a crash, please let me know.

batbrain7:
@arundhati24 I think testing the apk with this corresponding code in it is not possible for every PR , it is not a part of the review process, I asked you to check and tell if there are any crashes after this part in the review comments , so it is expected that you check the apk from your local branch and report back

arundhati24:
@batbrain7 I have already tested the apk from my local branch. And, there is no crash because of this. Rather there was an issue before making this change as this was leading to multiple calls to setRating(). I have tested the apks on my device and there is no crash till now.

PR #1488, opened 2018-07-03
Show toast only when the user rates a skill (Fixed #1454)

### LINKED ISSUE #1454: Toast pops up on opening skill details screen
**Actual Behaviour**

Currently, the toast 'Thank you for rating this skill' pops up even when the user has not rated the skill. This happens probably because the Toast has been called from inside the skill rating change listener. As the getRatingByUser.json API is called each time the skill details screen is visited, the rating on the rating bar changes due to which the toast pops up.

**Expected Behaviour**

The toast should only be displayed when the user rates a skill.

Yes

arundhati24:
Fixes #1454 

Changes: Now the toast pops up only when the user rates a skill and not each time the rating of the rating bar is updated.


arundhati24:

PR #1491, opened 2018-07-03
Replace multiple query params with query maps (Fixed #1431)

### LINKED ISSUE #1431: Replace multiple queries with query map
**Actual Behaviour**

Query map has been used to send query parameters during API calls. 

**Expected Behaviour**

Replace query map with object as it is not type safe.

Yes

arundhati24:
Fixes #1431 

Changes: Replaced multiple query params (more than three) with query maps in the final requests.


arundhati24:

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.java
```diff
@@ -229,8 +230,13 @@ private void rateSusiSkill(final String polarity, String locationUrl, final Cont
...
 
-        Call<SkillRatingResponse> call = new ClientBuilder().getSusiApi().rateSkill(susiLocation.get("model"),
-                susiLocation.get("group"), susiLocation.get("language"), susiLocation.get("skill"), polarity);
+        Map<String, String> map = new HashMap<>();
+        map.put("model", susiLocation.get("model"));
+        map.put("group", susiLocation.get("group"));
+        map.put("language", susiLocation.get("language"));
+        map.put("skill", susiLocation.get("skill"));
+        map.put("rating", polarity);
+        Call<SkillRatingResponse> call = new ClientBuilder().getSusiApi().rateSkill(map);
```
iamareebjamal: I said that use map once, in the service, and objects everywhere

This method is even more unsafe than the parameters because you can make mistake in the key name, the amount of keys and mismatch the key with value

Use the Map at the end of the service, just before the API call, and object everywhere else

batbrain7: @iamareebjamal a singleton instance ?

iamareebjamal: Yes

batbrain7: Okay making an issue for it.

arundhati24: @iamareebjamal This has been used just before the API call. Prior to this, there has been only one parameter 'polarity' that is being passed. In other cases like post feedback, I have used objects, as you had mentioned because there we have been passing more than three params and have used map in the final step i.e. just before the API call. As we need to send the URL string, I couldn't understand how to use objects in the final step to make an API call. Please have a look and if there is still a need to change something please let me know.

iamareebjamal: The map should not be created in ViewHolders and Views. It should be made in service

Wrap the API in a service and use it in the views

arundhati24: @iamareebjamal I am sorry but it would be really nice if you could please help me a bit with that. I don't understand completely how to do that. We already have [SusiService.java](https://github.com/fossasia/susi_android/blob/development/app/src/main/java/org/fossasia/susi/ai/rest/services/SusiService.java) where we wrap this API.
What modifications do I need to make to implement the changes that you requested?

iamareebjamal: You are creating Map thrice, in different places. I want it so that only one function creates the Map and takes object as a parameter, and all three places use that function and object instead of Map

arundhati24: @iamareebjamal So do I need to make a common function that receives a data class object, containing all required data, and returns a map?
Should this function be made in a separate helper class?

iamareebjamal: Ideally this should not return a Map, it should return a Call

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.java
```diff
@@ -229,8 +229,11 @@ private void rateSusiSkill(final String polarity, String locationUrl, final Cont
         if (susiLocation.size() == 0)
             return;
 
-        Call<SkillRatingResponse> call = new ClientBuilder().getSusiApi().rateSkill(susiLocation.get("model"),
-                susiLocation.get("group"), susiLocation.get("language"), susiLocation.get("skill"), polarity);
+        SkillRatingQuery queryObject = new SkillRatingQuery(susiLocation.get("model"), susiLocation.get("group"),
+                susiLocation.get("language"), susiLocation.get("skill"), polarity);
+        QueryHelper queryHelper = new QueryHelper();
```
iamareebjamal: Why is there a need to instantiate this?

This is basically a static method, besides the functions in this class should be placed in client builder

arundhati24:
@iamareebjamal @batbrain7 The build now passes. ClientBuilder.java was present in the debug folder and release folder but it showed unresolved reference for Play Store release as well as FDroid release variants while it was fine for the debug variants. When I moved the ClientBuilder.java to main folder, everything seemed to work fine. Please review this PR.

arundhati24:
@iamareebjamal @batbrain7 Please review this so that this can be merged before next release.

PR #1492, opened 2018-07-04
Fixed app crash on tapping examples (Fixed #1480)

### LINKED ISSUE #1480: Crash on tapping examples on skill details screen
**Actual Behaviour**

Currently, the app crashes sometimes on tapping examples on skill details screen due to null value for five star rating bar.

**Expected Behaviour**

The app should not crash.

**Logs**

screenshot_2018-07-04-12-49-29-215_com miui bugreport

Yes

arundhati24:
Fixes #1480 
Changes: Added null check for the five star rating bar. (Crash report can be seen in the linked issue)

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
@@ -532,4 +532,9 @@ class SkillDetailsFragment : Fragment(), ISkillDetailsView {
             }
         }
     }
+
+    override fun onDestroyView() {
+        skillDetailsPresenter.onDetach()
+        super.onDestroyView()
+    }
 }
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
@@ -53,11 +53,7 @@ class SkillDetailsFragment : Fragment(), ISkillDetailsView {
     private lateinit var skillTag: String
     private val imageLink = "https://raw.githubusercontent.com/fossasia/susi_skill_data/master/models/general/"
 
-    private lateinit var fiveStarSkillRatingBar: RatingBar
     private var fromUser = false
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
```diff
@@ -271,39 +271,41 @@ class SkillDetailsFragment : Fragment(), ISkillDetailsView {
         fiveStarSkillRatingBar = five_star_skill_rating_bar
         fiveStarSkillRatingScaleTextView = tv_five_star_skill_rating_scale
 
-        tvFiveStarSkillRatingBar.visibility = View.VISIBLE
-        fiveStarSkillRatingBar.visibility = View.VISIBLE
+        if (fiveStarSkillRatingBar != null) {
```
iamareebjamal: Rating Bar should never be null

Check why it is instead of adding a null check

batbrain7:
@arundhati24 this is an URGENT priority issue please work on this or else let me work on it ?

arundhati24:
@batbrain7 @iamareebjamal I am unable to understand why the rating bar becomes null sometimes. It doesn't happen every time. It leads to a crash once in a few trials.
@batbrain7 I also removed those extra variables this time. Removed lateinit variables and used direct ids instead, even now I got the same crash once which says fiveStarRatingBar must not be null.

batbrain7:
@arundhati24 could you help me reproduce the issue ?

batbrain7:
@arundhati24 I found the issue , the issue is with the skills that are already rated, since it takes some time for the rating bar to fetch that whether the user has already rated this skill or not and if the skill is rated the stars are updated and then we can click on examples to use them normally.
Also, in the mean time in which the fetching of the number of stars is done in that period of time if we click on the examples the app crashes otherwise it works fine.
So I suggest you to add some progressbar to show the user that rating is being fetched and then let the user interact with the skill details fragment.
Also @iamareebjamal can we do anything better ?

batbrain7:
Now it is a fairly easy job to do, please solve this quickly so that a new release can be made.

batbrain7:
@iamareebjamal I think @arundhati24 is still working on this

PR #1498, opened 2018-07-07
fix: the skills will be fetched always if internet connectivity is there

### LINKED ISSUE #1464: The pull to refresh is not disabled when skills are already being fetched.
**Actual Behaviour**
* The pull to refresh is not disabled when skills are already being fetched. It leads to multiple API calls and leads to request timeout. And the screen shows "unable to fetch data, please try again later"

**Expected Behaviour**
* The "pull to refresh"can be disabled while the interface is refreshing, so that the user can only refresh after the previous API call is completely handled.

**Steps to reproduce it**
Perform "Pull to Refresh" very quickly on the Skills Listing page.

**LogCat for the issue**
NA

**Screenshots of the issue**
NA

@arundhati24 , Can you work on it?

batbrain7:
Fixes #1464 

Changes: the UI was not getting updated once the "Unable to fetch data" was displayed , now the UI is updated so whenever we swipe down the UI is displayed properly.


batbrain7:
@iamareebjamal please merge

PR #1507, opened 2018-07-09
fix: the thumbs down option is shown in table response now

### LINKED ISSUE #1506: thumbs down was not shown in table response
**Actual Behaviour**

There is  no thumbs down button in the table response.

**Expected Behaviour**

There should be thumbs down button for the table type response.

Yes

batbrain7:
Fixes #1506 

@iamareebjamal @arundhati24 the variable name is changed and the thumbs down rating is shown now


app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
```diff
@@ -484,7 +484,7 @@ class ChatPresenter(chatActivity: ChatActivity) : IChatPresenter, IChatModel.OnR
                                 isHavingLink = psh.isHavingLink,
                                 datumList = psh.datumList,
                                 webSearch = psh.webSearch,
-                                tableData = tableDatas,
+                                tableData = tableItem,
```
iamareebjamal: `tableData` is still here

batbrain7:
@iamareebjamal the build is fixed now

PR #1511, opened 2018-07-10
fix: device section only visible on login

### LINKED ISSUE #1510: Display devices section only when the user is logged in
**Actual Behaviour**

Currently the devices section is displayed even when the user is not logged in.

**Expected Behaviour**

Hide the devices section when the user is not logged in

Yes

batbrain7:
Fixes #1510 

Changes: device section only visible when logged in


batbrain7:
@arundhati24 @iamareebjamal please review

app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
```diff
@@ -90,10 +90,14 @@ class ChatSettingsFragment : PreferenceFragmentCompat(), ISettingsView {
 
         // Display login email
         val utilModel = UtilModel(activity as SkillsActivity)
-        if (!utilModel.isLoggedIn())
+        if (!utilModel.isLoggedIn()) {
```
chiragw15: Shouldn't the Presenter be interacting with the model and not the view? @batbrain7

chiragw15: Shouldn't the Presenter be interacting with the model and not the view? @batbrain7

batbrain7: this code has been touched for the first time since last year @chiragw15 , much refactoring is needed, so this will be done in consequent issues or if it needs to be done now I'll refactor this Fragment and then update the PR ?

arundhati24: @batbrain7 Should it check for utilModel.isLoggedIn() or should it check for null acess token i.e. ```if(PrefManager.getToken() != null)``` for checking if the user is logged in because that automatically takes care of token expiry and also considers the case when user skips login?

batbrain7: thanks @arundhati24 I will use this instead

batbrain7:
@arundhati24 using the PrefManager now , if there is no other change please merge

batbrain7:
@iamareebjamal please merge

PR #1516, opened 2018-07-11
refactor: changed the device activity into fragments

### LINKED ISSUE #1489: To follow the Smart Speaker- Android connection workflow similar to Google Home
**Actual Behaviour**

No connection workflow right now

**Expected Behaviour**

To follow a workflow similar to Google Home

batbrain7:
Fixes #1516  

Moved the logic of devices activity to fragments , as we will have two different sections for devices one for the connection part and the other for the details part.
screenshot_20180712-200530  screenshot_20180712-201512  
screenshot_20180625-223742


batbrain7:
@arundhati24 @iamareebjamal the apk after this change is tested by @stealthanthrax and this still works fine even after converting to fragment

PR #1532, opened 2018-07-14
fix: app crash on susi button

### LINKED ISSUE #1531: App crash on clicking susi button on chat screen
**Actual Behaviour**

Whenever we click on susi button on the chat screen app crashes.

**Expected Behaviour**

App should not crash

**Steps to reproduce it**

Open the app then login then click on the susi button

**LogCat for the issue**

  ```
 Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'retrofit2.Call org.fossasia.susi.ai.rest.services.SusiService.fetchSkillMetricsData(java.lang.String, java.lang.String)' on a null object reference
        at org.fossasia.susi.ai.rest.ClientBuilder.fetchListSkillMetricsCall(ClientBuilder.java:98)
        at org.fossasia.susi.ai.data.SkillListingModel.fetchSkillMetrics(SkillListingModel.kt:61)
        at org.fossasia.susi.ai.skills.skilllisting.SkillListingPresenter.getMetrics(SkillListingPresenter.kt:47)
        at org.fossasia.susi.ai.skills.skilllisting.SkillListingFragment.onViewCreated(SkillListingFragment.kt:48)
```

Yes

### PR DESCRIPTION
**Pull request by batbrain7**
Fixes #1531 

Now since a singleton instance of retrofit is used, using the getter method for the `getSusiApi()` can be used to make the calls, and thus avoids the crash

### CODE REVIEW

File: app/src/main/java/org/fossasia/susi/ai/rest/ClientBuilder.java [hunk 1]
```diff
@@ -95,6 +95,8 @@ public static SusiService getSusiApi() {
     }
 
     public static Call<ListSkillMetricsResponse> fetchListSkillMetricsCall(SkillMetricsDataQuery queryObject) {
+        if (susiService == null)
+            createSusiService();
```
iamareebjamal: This issue is waiting to pop up again, rather than having `createSusiService`, add a getter which safely returns initialized service and use it everywhere

batbrain7: @iamareebjamal we already have getSusiApi() method, so maybe using that to return the calls should be fine

batbrain7: also @iamareebjamal I feel is there a need to put this Call in ClientBuilder when we can simply use

batbrain7:
@iamareebjamal I am sorry for this carelessness.

PR #1533, opened 2018-07-15
feat: make a basic device connection

batbrain7:
Fixes #1250 

This PR should not be reviewed on grounds for string resources and grammatical names as the strings will be deleted anyways, only the simplest connection possible is the aim of this PR.

A basic connection is established through the android app to the smart speaker. The API endpoints are hitting correctly, but due to recent code changes in the susi linux repository the code crashes on raspi, 
A follow up fixing PR has been sent for that by @stealthanthrax : https://github.com/fossasia/susi_linux/pull/258

here is the output  : 
8771327e-fd02-4808-820c-e23c8c11749f

PR #1537, opened 2018-07-15
 fix: receiver null error

batbrain7:
Fixes #1533 

Review changes from PR #1533 and null receiver error fix.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/device/deviceconnect/DeviceConnectPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/device/deviceconnect/DeviceConnectPresenter.kt
@@ -61,10 +61,10 @@ class DeviceConnectPresenter(deviceActivity: DeviceActivity, manager: WifiManage
         }
         if (!list.isEmpty()) {
             deviceConnectView?.setupWiFiAdapter(connections)
-//            deviceConnectView?.unregister()
+            // deviceConnectView?.unregister()
         } else {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/susi/ai/device/deviceconnect/DeviceConnectPresenter.kt
```diff
@@ -61,10 +61,10 @@ class DeviceConnectPresenter(deviceActivity: DeviceActivity, manager: WifiManage
         }
         if (!list.isEmpty()) {
             deviceConnectView?.setupWiFiAdapter(connections)
-            deviceConnectView?.unregister()
+//            deviceConnectView?.unregister()
         } else {
             deviceConnectView?.onDeviceConnectionError(utilModel.getString(R.string.no_device_found), utilModel.getString(R.string.setup_tut))
-            deviceConnectView?.unregister()
+//            deviceConnectView?.unregister()
```
codedsun: If not needed, remove the comment

batbrain7: It may be needed afterwards @codedsun

iamareebjamal: Then format the comments correctly acc. to Java style guidelines

Issue #1467, opened 2018-07-01
Skills listing UI flow needs modifications


arundhati24:
**Actual Behaviour**

Currently, there are a number of API calls being made in the skills listing activity as a separate API call is being made for each group. Also, when it is refreshed, there are a number of API calls, due to which there are a number of issues including crashed.

**Expected Behaviour**

Modify the skills listing UI flow to make it similar to CMS so as to reduce the number of API calls.

**Would you like to work on the issue?**

Yes

batbrain7:
@arundhati24 we have to do this urgently as a stable version would not be available until then

arundhati24:
Okay. I would be starting with it today itself then. Would put feedback on hold for sometime.

batbrain7:
Thanks :)

arundhati24:
This can be closed as all child issues have been completed.

PR #1538, opened 2018-07-16
feedback for the responses sent only once

### LINKED ISSUE #1526: Thumbs up or down cannot be changed
The feedback once given on an interaction should not be changeable, otherwise it will cause inconsistencies in the feedback logs that will be used for analysis.

No, anyone can assign the issue to himself/herself.

batbrain7:
Fixes #1526 

Changes: The feedback is now sent only once.

screenshot_20180718-124132 screenshot_20180718-124144


batbrain7:
@arundhati24 please review

batbrain7:
@arundhati24 review this PR , it is pending to be reviewed for 2 days now

arundhati24:
Yes, I have gone through this. Also, checked the issue. But, I still have a little doubt. What you want to do is to allow the user to rate the SUSI response only once. Is that so? And, what if the user wants to change his/her submitted rating?

batbrain7:
@arundhati24 in the issue it has been clearly stated by @AnupKumarPanwar that we cannot change the rating once given otherwise the feedback data given will be inconsistent, Also in the web client recently it has been added and google assistant also allows us to rate a skill only once.

arundhati24:
Okay, so once the user submits the rating, the rating bar consisting of thumbs up and thumbs down should be hidden?

batbrain7:
@arundhati no such thing is proposed yet , if it has to be hidden then it will done in a separate issue

arundhati24:
I think we should discuss it here and fix this in this PR because this is a complete thing to be done at once. How will the user know that s/he is allowed to rate a response only once? If we talk of Google Assistant, once the user submits the rating, the submitted rating is displayed on the screen and the option to rate skill is hidden. Here is the link to the [PR](https://github.com/fossasia/chat.susi.ai/pull/1453) on chat.susi.ai for the same issue. @AnupKumarPanwar @akshatnitd I think the UI change should also be done simultaneously, what are your views?

batbrain7:
@arundhati24 there is nothing wrong with adding the changes in the UI in this PR but a new user won't know this unless the app is updated so I would simply like to work on parent child issues as it makes the work pretty easy, similar to what you are doing in skills activity. @imjog will you do the same for iOS too?

arundhati24:
@batbrain7 The issues that I have opened implement one functionality at a time and not UI changes and associated backend code in separate PRs. For example you may have a look at #1467. 
Coming back to this issue, even the web client is following the similar approach. Plus, it's just about setting the visibility. Why can't it be done here?

batbrain7:
I don't follow you ? What do you mean by settings visibility?

akshatnitd:
I feel that, the UI changes should have been done first, as the user won't be able to rate the response again anyway in that case. And yes, apart from that, the changes in the functional aspect is important too. Try implementing the UI first, as the same is being done for web as well, as discussed in the analytics meeting. It has been a pending task and we want to quickly move ahead with the development, rather than splitting single issues and making the merging process slower. Will it be a major task to reflect changes on the UI as well? If I understand correctly, only a check needs to be added if it is already rated.

arundhati24:
@batbrain7 I'm sorry I didn't understand what you were trying to say. But I didn't write settings visibility. I just mentioned that setting the visibility of this view to gone would be required once the user has rated a response.

arundhati24:
I still opine the same as I already mentioned, but I don't think I would be able to comment any further on this. As usual, go ahead and implement it as you like.

batbrain7:
@arundhati24 the issue made does not provide any information about the UI changes, and it was not known to me that it was to be done , secondly why is following a parent child issue relationship not appropriate for this case?
Also the PR should contain the changes as defined in the issue only , in our case if the previous code needs to made better it should be done in the same PR too,
But here neither the issue defines the UI changes are to be made and nor there is any previous code to improved, Our goal is align web client and android client in terms of features and not in terms of PRs and issues.
It is not about a single check but is about being more organised and systematically clear, if still the UI changes need to be done in this PR I will do so ?

codedsun:
@batbrain7 @arundhati24 IMO, Instead of debating on PRs we should proceed with the work, if the issue needs too many changes and hiding the UI for thumbs up/down it should be done in a new PR. Their is no problem in opening issues/PR if they are relevant. Rest is all up your choice 👍

batbrain7:
Thanks @codedsun , though a new issue should be made but if @arundhati24 insists so much on changing the UI in this PR , I will do it in this only

batbrain7:
@arundhati24 @codedsun @akshatnitd you can review now

arundhati24:
@batbrain7 I am not insisting on anything. I already mentioned, as usual, proceed as you like. I just put forth my view and I might be wrong as well. I was just asking if that should be done. That's how things work. If I am saying something wrong, I should accept it and I will learn from it. But, that comes up only through discussions and reviews and that's how we learn. I don't have to insist on anything. Tell me what to do next, I will do as you say because I won't be able to comment any further on this.

batbrain7:
@arundhati24 your review would be much valuable here, Anup approved the changes for the functionality if you could look at them and approve if all seems good

batbrain7:
@akshatnitd the changes suggested have been done, please review

batbrain7:
@mariobehling @iamareebjamal please review and merge

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/TableViewHolder.java
```diff
@@ -100,55 +100,40 @@ public void setView(final ChatMessage model) {
...
-        }
+        thumbsDown.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                Timber.d(model.isPositiveRated() + " " + model.isNegativeRated());
+                if (!model.isPositiveRated() && !model.isNegativeRated()) {
+                    thumbsDown.setImageResource(R.drawable.thumbs_down_solid);
+                    rateSusiSkill(Constant.NEGATIVE, model.getSkillLocation(), itemView.getContext());
+                    setRating(true, false);
+                }
```
iamareebjamal: This is used everywhere, create a function and reuse

batbrain7: @iamareebjamal make a function for this in each file ?

iamareebjamal: No, make a static function

batbrain7: @iamareebjamal each viewholder has different view id as different layouts , so where should I place the common function?

iamareebjamal: What's same?

batbrain7: actually I don't think there is anything common among the different view holders, but in each viewholder this code is written twice so a method for this can be made in each viewholder

iamareebjamal: OK, let's start with that

batbrain7: private void setThumbsUpThumbsDown(boolean rating) {
        if (!model.isPositiveRated() && !model.isNegativeRated()) {
            if (rating) {
                thumbsUp.setImageResource(R.drawable.thumbs_up_solid);
                rateSusiSkill(Constant.POSITIVE, model.getSkillLocation(), itemView.getContext());
                setRating(true, true);
            } else {
                thumbsDown.setImageResource(R.drawable.thumbs_down_solid);
                rateSusiSkill(Constant.NEGATIVE, model.getSkillLocation(), itemView.getContext());
                setRating(true, false);
            }
        }
    }

iamareebjamal: I'm seeing a lot of duplicate code, I don't know how you cannot see it

I'll look at it and let you know what to do

PR #1556, opened 2018-07-18
 fix: app does not accept device responses

### LINKED ISSUE #1555: Make the API calls to speaker accept responses
**Actual Behaviour**

Currently when the API calls are made to the speaker the app does accept any response from the local server and so the error was caught in the error listener everytime.

**Expected Behaviour**

The app should accept response from the local server.

**Steps to reproduce it**

Setup your device and then try to make connection with the android app.

Yes

batbrain7:
Fixes #1555 

Added data serialisable classes for each api type to accept responses from the local server.

Screenshots for the change: 
c6c5193e-d27a-4abd-9868-7c6646ecb1e1


app/src/main/java/org/fossasia/susi/ai/data/DeviceModel.kt
```diff
@@ -26,16 +29,21 @@ class DeviceModel : IDeviceModel {
...
             }
 
-            override fun onResponse(call: Call<String>?, response: Response<String>?) {
+            override fun onResponse(call: Call<SpeakerWifiResponse>, response: Response<SpeakerWifiResponse>) {
                 listener.onSendCredentialSuccess()
             }
         })
     }
 
     override fun setConfiguration(stt: String, tts: String, hotword: String, wake: String, listener: IDeviceModel.onSetConfigurationListener) {
```
iamareebjamal: Convert to object

batbrain7:
@iamareebjamal could you please review the changes in this PR

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerAuthResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+    var auth: String? = null
+
+    @SerializedName("authentication")
+    var authentication: String? = null
+
+    @SerializedName("email")
+    var email: String? = null
+
+    @SerializedName("password")
+    var password: String? = null
```
iamareebjamal: Why not val, and SerializedName is not needed when variable name is same as field name

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerConfigResponse.kt
```diff
@@ -0,0 +1,21 @@
...
+    var hotword: String? = null
+
+    @SerializedName("stt")
+    var stt: String? = null
+
+    @SerializedName("tts")
+    var tts: String? = null
+
+    @SerializedName("wake")
+    var wake: String? = null
```
iamareebjamal: Same

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerConfigResponse.kt
```diff
@@ -0,0 +1,14 @@
+package org.fossasia.susi.ai.rest.responses.others
+
+class SpeakerConfigResponse {
+
```
arundhati24: Can this be converted to data class?

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerAuthResponse.kt
```diff
@@ -0,0 +1,12 @@
+package org.fossasia.susi.ai.rest.responses.others
+
+class SpeakerAuthResponse {
+
```
arundhati24: Can this be converted to data class?

app/src/main/java/org/fossasia/susi/ai/data/DeviceModel.kt
```diff
@@ -26,16 +30,21 @@ class DeviceModel : IDeviceModel {
...
         })
     }
 
-    override fun setConfiguration(stt: String, tts: String, hotword: String, wake: String, listener: IDeviceModel.onSetConfigurationListener) {
+    override fun setConfiguration(speakerConfig: SpeakerConfiguration, listener: IDeviceModel.onSetConfigurationListener) {
+        val query: MutableMap<String, String> = HashMap()
+        query.put("stt", speakerConfig.stt)
+        query.put("tts", speakerConfig.tts)
+        query.put("hotword", speakerConfig.hotword)
+        query.put("wake", speakerConfig.wake)
```
arundhati24: You can make a static method in ClientBuilder.java where you can create this map there and return a call from there.

batbrain7: ClientBuilder is only for api calls to susi server, but these calls are made to the local server of the device.

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerConfigResponse.kt
batbrain7: I think yes, but what difference does it make , all the response classes were made like this so made this the same way as well

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerAuthResponse.kt
batbrain7: same as above

app/src/main/java/org/fossasia/susi/ai/data/DeviceModel.kt
batbrain7: @iamareebjamal there is no clientbuilder for device service should this be changed?

app/src/main/java/org/fossasia/susi/ai/rest/responses/others/SpeakerConfigResponse.kt
iamareebjamal: Well, we should not make the mistakes previous developers made. Plus those classes were converted from Java to Kotlin, which does not automatically change them to data classes

`but what difference does it make`
Also, read up why data classes are used and what's the difference from normal classes, you'll know why I'm telling you to use data classes

app/src/main/java/org/fossasia/susi/ai/data/DeviceModel.kt
iamareebjamal: Ideally there should be a wrapper over retrofit services which does this stuff and validation, processing before return etc

But previous work which @arundhati24 worked on was already built and used, so it was a huge change to be done there. This is new implementation, add a service on top which internally calls the API

iamareebjamal:
@batbrain7 That should handle other classes, why should we add a new class in wrong format to just correct it again?

batbrain7:
@iamareebjamal could you check the implementation for the service that calls the API ?

batbrain7:
Thanks @iamareebjamal it would be great if this PR gets merged before the next release

batbrain7:
@iamareebjamal after the changes the app is unable to hit the API points or is unable to catch the listener event

iamareebjamal:
Check what request is generated and sent and what is the reponse

batbrain7:
it works @iamareebjamal , there was issue from server please review and merge this PR

PR #1563, opened 2018-07-20
Fixed problem loading skills when the user has not set a language explicitly (Fixed #1485)

### LINKED ISSUE #1485: Skills not loaded when a user has not set a language explicitly
**Actual Behaviour**

Currently, if a user has not set a language explicitly from the settings at least once in out of the entire duration of app usage, the skills are not loaded because the response returned from the server contains default value which is not present in the client, where en is being considered as default language.

**Expected Behaviour**

Show skills in any case.

**Steps to reproduce**

Log into SUSI from an account for which you have never changed language from Settings. Then go to the skills listing page. You will not see the skills loaded. Only groups list can be seen.

Yes

arundhati24:
Fixes #1485 

Changes: Now the skills are loaded even when the user logs in for the first time.


arundhati24:
No, there are steps to reproduce. You can try reproducing it.

PR #1577, opened 2018-07-21
fix: scan to device shown

### LINKED ISSUE #1570: Scan for devices shown in background
**Actual Behaviour**

whenever we click on click to setup device and the location is off we see Scanning for devices in the background

**Expected Behaviour**

This should not be seen.

**Steps to reproduce it**

login the app, click on susi button , open settings, click on device setup

Yes

batbrain7:
Fixes #1570 

Changes: Now it is not shown in the background and only shown devices are scanned

Screenshots for the change: 
screenshot_20180721-104324 screenshot_20180721-104107


batbrain7:
@iamareebjamal @mariobehling please merge

PR #1614, opened 2018-07-31
fix: crash in fdroid flavor

### LINKED ISSUE #1613: crashes in fdroid flavor
**Actual Behaviour**

Since the fdroid flavor has the version in which the SUSI.AI app will not have the hotword detection feature, there are certain crashes which must be handled that occur because of that.

**Expected Behaviour**

make the fdroid flavor crash free after the prebuilt files are removed.

Yes

batbrain7:
Fixes #1613

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -34,6 +34,7 @@ import android.view.inputmethod.InputMethodManager
 import android.widget.Toast
 import io.realm.RealmResults
 import kotlinx.android.synthetic.main.activity_chat.*
+import org.fossasia.susi.ai.BuildConfig
 import org.fossasia.susi.ai.R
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/chat/ChatPresenter.kt
```diff
@@ -96,7 +97,7 @@ class ChatPresenter(chatActivity: ChatActivity) : IChatPresenter, IChatModel.OnR
     //initiates hotword detection
     override fun initiateHotwordDetection() {
         if (chatView!!.checkPermission(utilModel.permissionsToGet()[2]) &&
-                chatView!!.checkPermission(utilModel.permissionsToGet()[1])) {
+                chatView!!.checkPermission(utilModel.permissionsToGet()[1]) && BuildConfig.FLAVOR.equals("playStore")) {
```
iamareebjamal: Do early return

batbrain7:
@iamareebjamal merge this, now after this merge and building susi again with fdroid we can send the merge request

PR #1620, opened 2018-08-01
Anonymize feedback and add unit tests for the method added to the UtilModel (Fixed #1615)

### LINKED ISSUE #1615: Anonymize feedback section
**Actual Behaviour**

The feedback section currently shows complete email addresses. 

**Expected Behaviour**

Anonymize the feedback and only show emails until the @ sign, e.g. "abc@..".

Yes

arundhati24:
Fixes #1615 

Changes: Now the complete email is not shown. Rather, an email is truncated after '@' symbol.

**Screenshot**

truncateemail

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/helper/Utils.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/helper/Utils.kt
@@ -60,4 +60,14 @@ object Utils {
         return result.toString()
     }
 
+    fun truncateEmailAtEnd(email: String?): String? {
+        if (!email.isNullOrEmpty()) {
+            val truncateAt = email?.indexOf('@')
+            if (truncateAt != null && truncateAt != -1) {
+                return email.substring(0, truncateAt.plus(1)) + " ..."
+            }
+        }
+        return null
+    }
+
 }
--- a/app/src/main/java/org/fossasia/susi/ai/skills/feedback/adapters/recycleradapters/AllReviewsAdapter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/feedback/adapters/recycleradapters/AllReviewsAdapter.kt
@@ -6,9 +6,7 @@ import android.support.v7.widget.RecyclerView
 import android.text.TextUtils
 import android.view.LayoutInflater
 import android.view.ViewGroup
-import com.squareup.picasso.Picasso
 import org.fossasia.susi.ai.R
-import org.fossasia.susi.ai.helper.CircleTransform
 import org.fossasia.susi.ai.helper.Utils
 import org.fossasia.susi.ai.rest.responses.susi.Feedback
 import org.fossasia.susi.ai.skills.feedback.adapters.viewholders.AllReviewsViewHolder
@@ -42,7 +40,7 @@ class AllReviewsAdapter(val context: Context, val feedbackList: List<Feedback>?)
                 if (feedbackList[position].email != null &&
                         !TextUtils.isEmpty(feedbackList[position].email)) {
                     Utils.setAvatar(context, feedbackList.get(position).email, holder.avatar)
-                    holder.feedbackEmail?.text = feedbackList[position].email
+                    Utils.truncateEmailAtEnd(feedbackList.get(position).email)?.let { holder.feedbackEmail?.text = it }
                 }
                 if (feedbackList[position].timestamp != null &&
                         !TextUtils.isEmpty(feedbackList[position].timestamp)) {
--- a/app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
+++ b/app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
@@ -1,12 +1,25 @@
 package org.fossasia.susi.ai.helper
 
 import junit.framework.Assert.assertEquals
-import junit.framework.Assert.assertTrue
-import org.fossasia.susi.ai.rest.responses.susi.SkillData
-import org.junit.Before
 import org.junit.Test
+import org.fossasia.susi.ai.rest.responses.susi.SkillData
 
+/**
+ *
+ * Created by arundhati24 on 03/08/2018.
+ */
 class UtilsTest {
+    @Test
+    fun testTruncateEmailAtEnd() {
+        assertEquals("testuser@ ...", Utils.truncateEmailAtEnd("testuser@example.com"))
+        assertEquals(null, Utils.truncateEmailAtEnd("testuser"))
+        assertEquals(null, Utils.truncateEmailAtEnd(""))
+        assertEquals(null, Utils.truncateEmailAtEnd(" "))
+        assertEquals(null, Utils.truncateEmailAtEnd(null))
+        assertEquals("test.user@ ...", Utils.truncateEmailAtEnd("test.user@example.com"))
+        assertEquals("test_user@ ...", Utils.truncateEmailAtEnd("test_user@example.com"))
+        assertEquals("test123@ ...", Utils.truncateEmailAtEnd("test123@example.com"))
+    }
 
     @Test
     fun verifyImageLink() {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -117,6 +117,8 @@ dependencies {
     implementation "com.takisoft.fix:preference-v7:${rootConfiguration.preferenceVersion}"
 
     //unit test
+    testImplementation "junit:junit:${rootConfiguration.jUnitVersion}"
+    testImplementation "org.mockito:mockito-core:${rootConfiguration.mockitoCoreVersion}"
     androidTestImplementation "com.android.support:support-annotations:${rootConfiguration.supportLibraryVersion}"
     androidTestImplementation "com.android.support.test:runner:${rootConfiguration.testVersion}"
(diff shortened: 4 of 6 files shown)


app/src/main/java/org/fossasia/susi/ai/skills/feedback/adapters/recycleradapters/AllReviewsAdapter.kt
```diff
@@ -39,7 +39,10 @@ class AllReviewsAdapter(val context: Context, val feedbackList: List<Feedback>?)
                 if (feedbackList[position].email != null &&
                         !TextUtils.isEmpty(feedbackList[position].email)) {
                     holder.initials.text = feedbackList[position].email?.substring(0, 2)
-                    holder.feedbackEmail?.text = feedbackList[position].email
+                    val truncateAt = feedbackList.get(position).email?.indexOf('@')
+                    if (truncateAt != null && truncateAt is Int) {
+                        holder.feedbackEmail?.text = feedbackList.get(position).email?.substring(0, truncateAt.plus(1)) + " ..."
+                    }
```
iamareebjamal: Add the method in Utils as this is a helper method reused in both files

Duplication is bad

Also, add tests

arundhati24:
@iamareebjamal I made the changes and also added test for the added method. Please have a look.

app/src/main/java/org/fossasia/susi/ai/data/UtilModel.kt
```diff
@@ -17,6 +17,15 @@ import retrofit2.Response
...
 
+    companion object {
+        fun truncateEmailAtEnd(email: String?): String? {
+            val truncateAt = email?.indexOf('@')
+            if (truncateAt is Int) {
+                return (email.substring(0, truncateAt.plus(1)) + " ...")
+            }
+            return null
+        }
+    }
```
iamareebjamal: Should be in Utils object

app/src/test/java/data/UtilModelTest.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Created by arundhati24 on 03/08/2018.
+ */
+class UtilModelTest {
+
+    @Test
+    fun testTruncateEmailAtEnd() {
+        val actualEmail = UtilModel.truncateEmailAtEnd("testuser@example.com")
+        val expectedEmail = "testuser@ ..."
+        assert(actualEmail?.trim().equals(expectedEmail))
```
iamareebjamal: Don't trim and use `assertEquals`

iamareebjamal: Add 2 3 more tests for edge cases where the test might fail

app/src/test/java/helper/UtilsTest.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.Test
+
+/**
+ *
+ * Created by arundhati24 on 03/08/2018.
+ */
+class UtilsTest {
+    @Test
+    fun testTruncateEmailAtEnd() {
+        assertEquals(true, Utils.truncateEmailAtEnd("testuser@example.com").equals("testuser@ ..."))
```
iamareebjamal: Wrong use of assetEquals

app/src/test/java/helper/UtilsTest.kt
```diff
@@ -0,0 +1,19 @@
+package helper
```
iamareebjamal: Use the same package as the file to be tested

iamareebjamal:
@arundhati24 In how many days will this be fixed?

arundhati24:
@iamareebjamal It will be fixed today itself.

app/src/test/java/helper/UtilsTest.kt
arundhati24: @iamareebjamal assertEquals() takes two params expected value and actual value, so is it because of the passed params?

iamareebjamal: `truncateEmailAtEnd` returns String and not boolean. So expected param should be?

arundhati24:
@iamareebjamal Please review this PR. (Codacy only shows two documentation issues, which need to be ignored I think).

app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.junit.Test
+
+/**
+ *
+ * Created by arundhati24 on 03/08/2018.
+ */
+class UtilsTest {
+    @Test
+    fun testTruncateEmailAtEnd() {
+        assertEquals("testuser@ ...", Utils.truncateEmailAtEnd("testuser@example.com"))
```
iamareebjamal: Add more emails which may produce errors. Like emails with period and special symbols. Goad of tests is to break the code as much it is to test if it is working correctly

app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
```diff
@@ -0,0 +1,18 @@
...
+        assertEquals(null, Utils.truncateEmailAtEnd("testuser"))
+        assertEquals(null, Utils.truncateEmailAtEnd(""))
+        assertEquals(null, Utils.truncateEmailAtEnd(" "))
```
iamareebjamal: Add test for null

app/src/main/java/org/fossasia/susi/ai/helper/Utils.kt
```diff
@@ -58,4 +57,13 @@ object Utils {
         return result.toString()
     }
 
+    fun truncateEmailAtEnd(email: String?): String? {
+        if (!email.isNullOrEmpty()) {
+            val truncateAt = email?.indexOf('@')
+            if (truncateAt is Int && truncateAt != -1) {
```
iamareebjamal: `truncateAt is Int` is not required?

app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
iamareebjamal: What about emails with multiple `@` symbols?

app/src/main/java/org/fossasia/susi/ai/helper/Utils.kt
```diff
@@ -58,4 +57,13 @@ object Utils {
...
+                return (email.substring(0, truncateAt.plus(1)) + " ...")
```
iamareebjamal: Extra paranthesis

app/src/main/java/org/fossasia/susi/ai/helper/Utils.kt
arundhati24: As email is nullable, ```truncateAt``` is being called upon a nullable String object. Later when I have used ```truncateAt.plus(1)``` for that we strictly require ```Int``` but ```truncateAt``` is of type ```Int?```
Hence, I have used ```truncateAt is Int``` when the email is not null or empty.
Could this be improved?

iamareebjamal: `truncateAt != null and ...`

app/src/test/java/org/fossasia/susi/ai/helper/UtilsTest.kt
arundhati24: @iamareebjamal As this type of email is invalid, hence null should be returned. To check for this, I tried checking for the validity of the email in the ```truncateEmailAtEnd()``` using the ```CredentialsHelper.kt``` class which in turn uses the TextUtils class to check for the validity of the email. But, the tests fail with the error ```java.lang.RuntimeException: Method isEmpty in android.text.TextUtils not mocked. See http://g.co/androidstudio/not-mocked for details```

On checking the link, I found this :

screen shot 2018-08-10 at 8 04 51 pm

On adding the above code snippet to build.gradle and then running the tests I received an NPE (though I had checked for nullability of email and then used !!)

```
fun truncateEmailAtEnd(email: String?): String? {
        if (!email.isNullOrEmpty()) {
            if (CredentialHelper.isEmailValid(email!!)) {
                val truncateAt = email.indexOf('@')
                if (truncateAt is Int && truncateAt != -1) {
                    return email.substring(0, truncateAt.plus(1)) + " ..."
                }
            }
        }
        return null
    }
```

iamareebjamal: You cannot use android classes in java

arundhati24: Would you please suggest how should I account for the cases of multiple '@' or '@' followed by any random symbols?

iamareebjamal: I think that it should simply be truncated after a threshold amount of characters, not according to the position of `@` symbol

arundhati24: @iamareebjamal I feel what you have mentioned is true. But, as this is how this has been done on Skills CMS and iOS to anonymise the identity of the user, I need to implement it in the same way in the Android app.
Hence, please suggest what I should do in order to add tests for such a case (although an invalid email is not expected as the response is coming from the server).

PR #1629, opened 2018-08-01
Remove unused susi response file (Fixed #1627)

### LINKED ISSUE #1627: Unused susi response file
**Actual Behaviour**

There is an unused response file called 'Skills'.

**Expected Behaviour**

Remove unused file.

Yes

arundhati24:
Fixes #1627


batbrain7:
This api maybe used in the future, we never when the flow can change, removing this is not such a good option

arundhati24:
This is not an API. This is a response file. For your information, this will not be used. Instead filteredData array is used everywhere now. Hence, this is not needed.

arundhati24:
@batbrain7 I resolved the conflicts.
@iamareebjamal As this PR is leading to unnecessary conflicts in every PR, can this be merged if it is fine?

iamareebjamal:
@batbrain7 Please don't use Request Changes for asking for conflict resolution because then I have to dismiss the review. Github automatically doesn't allow merging until conflicts are resolved, just use a single comment

PR #1634, opened 2018-08-01
Fix:1631 Empty Feedback not sent

### LINKED ISSUE #1631: Skill Feedback Sent on Empty String
**Actual Behaviour**

Skill feedback sent on empty or whitespace string 

**Expected Behaviour**

It should not be sent

**Steps to reproduce it**

Go to a skill and post an whitespace feedback

Yes

codedsun:
Fixes #1631 : Empty Feedback is not send

Changes: Trim method applied on string.


iamareebjamal:
I have no idea why don't you follow the commit guidelines @codedsun

PR #1654, opened 2018-08-09
Add horizontal recyclerview to display staff picks on skills listing page (Fixed #1653)

### LINKED ISSUE #1653: Add horizontal recyclerview for staff picks
Add a horizontal recycler view to display top ten skills under the category staff picks in the skills listing fragment.
 
Yes

arundhati24:
Fixes #1653 

Screenshots for the change: 

staffpicks

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
@@ -98,6 +98,15 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
             if (metricsData != null) {
                 metrics.metricsList.clear()
                 metrics.metricsGroupTitles.clear()
+                if (metricsData?.staffPicks != null && metricsData?.staffPicks?.size != null) {
+                    //TODO : Make this comparison null safe
+                    if (metricsData?.staffPicks?.size!! > 0) {
+                        metrics.metricsGroupTitles.add(utilModel.getString(R.string.metric_staff_picks))
+                        metrics.metricsList.add(metricsData?.staffPicks)
+                        skillListingView?.updateAdapter(metrics)
+                    }
+                }
+
                 if (metricsData?.rating != null) {
                     if (metricsData?.rating?.size as Int > 0) {
                         metrics.metricsGroupTitles.add(utilModel.getString(R.string.metric_rating))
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
```diff
@@ -98,6 +98,14 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
             if (metricsData != null) {
                 metrics.metricsList.clear()
                 metrics.metricsGroupTitles.clear()
+                if (metricsData?.staffPicks != null) {
+                    if (metricsData?.staffPicks?.size as Int > 0) {
```
iamareebjamal: `if (metricsData?.staffPicks != null && metricsData?.staffPicks?.size as Int > 0) {`

Remove above condition

app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
```diff
@@ -98,6 +98,12 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
             if (metricsData != null) {
                 metrics.metricsList.clear()
                 metrics.metricsGroupTitles.clear()
+                if (metricsData?.staffPicks != null && metricsData?.staffPicks?.size as Int > 0) {
```
iamareebjamal: Remove `as Int`

arundhati24: @iamareebjamal As this is a nullable receiver, Android Studio shows a warning that this call corresponds to a dot qualified call, which is not allowed on a nullable receiver.
One way to fix this is to declare a local val, say 'size',  and then cast size to 
Int using ```if (size is Int) ``` and then checking for ```size > 0``` So, should I change this?

iamareebjamal: `metricsData?.staffPicks?.size != null && metricsData?.staffPicks?.size > 0`

arundhati24: 'Smart cast to 'Int' is impossible because ```metricsData?.staffPicks?.size``` is a complex expression.'
This can be overcome by assigning this to a separate var, say size, and then applying safe cast on size and checking for the condition.

arundhati24: 'staffPicks' is an Array. Even if I convert it 'toTypedArray()', it will still need a cast because of a nullable receiver. The other option is to use !! after checking if ```metricsData?.staffPicks != null``` which is not safe to use.

iamareebjamal: > The other option is to use !! after checking if metricsData?.staffPicks != null which is not safe to use.

Your statement is self-contradictory. It is definitely safe if you have checked that it is not null. Use !! but add a todo statement to make null safe

PR #1684, opened 2018-08-29
fix: App crash on opening settings activity

### LINKED ISSUE #1682: App Crash on opening settings
**Actual Behaviour**

The app is crashing on opening settings activity.

**Expected Behaviour**

It should not crash

**Steps to reproduce it**

Open the app, step into settings activity

**LogCat for the issue**

```
08-29 19:48:16.088 19310-19310/ai.susi E/AndroidRuntime: FATAL EXCEPTION: main
    Process: ai.susi, PID: 19310
    java.lang.IllegalStateException: preferenceManager.findPr…rence(Constant.MIC_INPUT) must not be null
        at org.fossasia.susi.ai.skills.settings.ChatSettingsFragment.onCreatePreferencesFix(ChatSettingsFragment.kt:80)
        at com.takisoft.fix.support.v7.preference.PreferenceFragmentCompat.onCreate(PreferenceFragmentCompat.java:62)
        at android.support.v4.app.Fragment.performCreate(Fragment.java:2331)
        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1386)
        at android.support.v4.app.FragmentTransition.addToFirstInLastOut(FragmentTransition.java:1188)
        at android.support.v4.app.FragmentTransition.calculateFragments(FragmentTransition.java:1071)
        at android.support.v4.app.FragmentTransition.startTransitions(FragmentTransition.java:115)
        at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2380)
        at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2338)
        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2245)
        at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:703)
        at android.os.Handler.handleCallback(Handler.java:836)
        at android.os.Handler.dispatchMessage(Handler.java:103)
        at android.os.Looper.loop(Looper.java:208)
        at android.app.ActivityThread.main(ActivityThread.java:6267)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1063)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:924)

```

**Screenshots of the issue**

ANR Dialog!

Yes!

codedsun:
Fixes #1682 

Changes: Changed the key of preference screen.

@arundhati24 @batbrain7 Check!


codedsun:
@batbrain7 @arundhati24 Check this!

codedsun:
@arundhati24 @batbrain7 Check this!

app/src/main/res/values/strings.xml
```diff
@@ -78,18 +78,18 @@
 
     <string name="setting_mic_input">Mic input</string>
     <string name="setting_hotword_detection">Hotword Detection (Beta)</string>
-    <string name="setting_mic_key" translatable="false">micInput</string>
+    <string name="setting_mic_key" translatable="false">mic_input</string>
```
arundhati24: @iamareebjamal I saw translations of these strings in other localization files, which I think should not be there because the Lint tool does not need to look for its translation in any of the res files.
If 'false' has been assigned to the translatable attribute of a string resource, is there a need to add the translations to these strings in other localization files?

iamareebjamal: No

iamareebjamal:
Rather than just changing the keys, don't use Constants file at all and load everything from strings

Duplication is evil

app/src/main/res/values/strings.xml
arundhati24: 👍 Will be removing unnecessary translations.

codedsun:
@arundhati24 What shall be done for this PR?

arundhati24:
@codedsun  Please check the comment from @iamareebjamal and work upon it.

codedsun:
@arundhati24 What I can understand to not use Constant file and use string resource as it's everywhere?

arundhati24:
@codedsun Yes. This is what is needed.

codedsun:
@arundhati24 Would like to ask , that constants file is used more than 100+ in project. What has to be done? I have removed the unnecssary translations. Please guide me

codedsun:
@arundhati24 I would request to state, what duplications are being talked about so that I can change that and get this PR ready asap!

Because string resources are used as key in preference screen while constants file values are used during checking the values in presenter? Shall that be replaced

` micCheck = utilModel.getBooleanPref(Constant.MIC_INPUT, true)`

batbrain7:
@codedsun please resolve conflicts

codedsun:
@batbrain7 Please read the problem I am facing!

codedsun:
@arundhati24 Please guide me on this too, I have changed few files

batbrain7:
@codedsun you just have to check the duplicated strings, which have a fixed value and the value of which is not seen on the screen as it is irrespective of the language.

codedsun:
@batbrain7 I think, I did the changes please chck the PR

codedsun:
@batbrain7 I have kept all the values in values/strings.xml and removed unnecessary translations in every other files.

arundhati24:
@codedsun First of all update your branch. Further, what @iamareebjamal said was to remove duplication i.e. if you see closely, you will find that in some places you are using "enter_send" key using the strings.xml file while at others you are using the Constants.java file to refer to the same key. For e.g. check the constant for "enter_send", you will find it in both the files. So, all he said was to refer to these keys using just one source i.e. the strings.xml file, and discontinue the use of Constants file. This would be better because in any case if you ever have to, say, change the key, you just need to do it at one place. If you have such duplications, there might be unnecessary issue arising which might take unnecessary time and attention.

codedsun:
@arundhati24 Please take a look and let me know if their's something else to be done! Lest, I have done the changes!

app/src/main/java/org/fossasia/susi/ai/helper/PrefManager.java
```diff
@@ -237,17 +238,18 @@ public static void clearToken() {
      *
      * @return the boolean
      */
-    public static boolean checkSpeechOutputPref() {
-        return PrefManager.getBoolean(Constant.SPEECH_OUTPUT, true);
+    public static boolean checkSpeechOutputPref(Context context) {
+//        return PrefManager.getBoolean(Constant.SPEECH_OUTPUT, true);
```
arundhati24: This is extra.

arundhati24:
@codedsun Please check the review from @iamareebjamal and make the changes requested, we need to fix this before updating the app on Play Store.

codedsun:
@arundhati24 Review!

codedsun:
@iamareebjamal I am not able to get the context through PrefManager class, either we should go by defining the method to get the context using MainApplication.getInstance( ) or we can define a method which returns the boolean as done for checkSpeechOutput.

iamareebjamal:
There is already sharedpreferences in pref manager, and you can't get it without context, so... recheck your claim that you can't get context in pref manager

codedsun:
@iamareebjamal They have used MainApplication.getInstance method to get the context that is used to get the shared preferences. How can I get it using the PrefManager class, i think I am not able to figure it out.

iamareebjamal:
You have answered your question yourself. I don't know what else to explain. @arundhati24 can you please help?

codedsun:
Replacing util.getString to MainApplication.getInstance.baseContext.?

codedsun:
What if I can define a method to get the context from the PrefManager and then can use it anywhere, and can also save lots of changes inside the implementation of get/put methods.

Shall i go with this approach?

iamareebjamal:
No, we want to reduce the complexity, not increase it.

Instead of `PrefManager.putBoolean(utilModel.getString(R.string.blah), true);`, you are saying that we use `PrefManager.putBoolean(PrefManager.getContext().getString(R.string.blah), true);`

Which is in no way better at all, I'd even call it worse because we are getting context from PreferenceManager, it means it already does have context, why does it need for us to call getString on it?

I'm asking you to change this so that each call looks like this

`PrefManager.putBoolean(R.string.blah, true);`

Which is extremely better than your current and suggested approach. We must follow the path of least resistance, whereas your approach fights the framework even more. About the amount of changes, it would have taken less work if you started with this pattern from the beginning instead of using utilModel to get string and passing in PreferenceManager

codedsun:
@iamareebjamal Did the changes, please check!

app/src/main/java/org/fossasia/susi/ai/data/contract/IUtilModel.kt
```diff
@@ -20,8 +20,8 @@ interface IUtilModel {
     fun setServer(isSusiServer: Boolean)
     fun setCustomURL(url: String)
     fun getString(id: Int): String
-    fun getBooleanPref(prefName: String, defaultValue: Boolean): Boolean
-    fun putBooleanPref(prefName: String, value: Boolean)
+    fun getBooleanPref(prefName: Int, defaultValue: Boolean): Boolean
+    fun putBooleanPref(prefName: Int, value: Boolean)
```
iamareebjamal: Good job, instead of replacing these methods, just add two methods here

codedsun: @iamareebjamal umm, but I have changed their implementation and usage everywhere! and their are still some files using from CONSTANTS, would create an issue and send a PR!

iamareebjamal: You just have to add two methods here, no need to change any other file

arundhati24:
@codedsun Please check reviews from @iamareebjamal and it would be great if you could please finalise this PR. This crash needs to be fixed soon.

codedsun:
@arundhati24 Finalized!

codedsun:
@iamareebjamal Please check!

PR #1709, opened 2018-09-26
moved lambda expression inside parentheses outside of it

### LINKED ISSUE #1708: Lambda argument should be moved out of parentheses
**Actual Behaviour**
Inside signup activity lambda expression inside parentheses which can be moved outside of them.

Yes

rakshitjadon:
Fixes #1708


arundhati24:
@rakshitjadon Did you get a chance to check the review from @iamareebjamal ? Did you find any more such instances? This indeed is a very minor change for a PR.

PR #1714, opened 2018-09-26
fix: Fix multiple spawn of SpeechInputlistner

### LINKED ISSUE #1713: App crash due to multiple spawning of voice input
**Actual Behaviour**

The fragment which is called when voice input is pressed does not disable the voice input and thus if the user presses on the same location more than once multiple voice input is activated and if it's more than twice then the app crashes

**Expected Behaviour**

The voice input should be disabled for the entire duration until the input is given to susi

**Steps to reproduce it**
Press on the voice input button multiple times (even if it's not visible)

Yes

kartik-karz:
Fixes #1713

Changes: Voice input is disabled until the previous input is running

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -287,6 +287,11 @@ class ChatActivity : AppCompatActivity(), IChatView {
 
     }
 
+    fun enableVoiceInput() {
+        btnSpeak.setImageResource(R.drawable.ic_mic_24dp)
+        btnSpeak.isEnabled = true
+    }
+
     override fun showWaitingDots() {
         recyclerAdapter.showDots()
     }
@@ -341,6 +346,7 @@ class ChatActivity : AppCompatActivity(), IChatView {
             chatPresenter.check(true)
             btnSpeak.setImageResource(R.drawable.ic_mic_24dp)
             btnSpeak.setOnClickListener({
+                btnSpeak.isEnabled = false
                 textToSpeech?.stop()
                 chatPresenter.startSpeechInput()
             })
--- a/app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
@@ -122,7 +122,10 @@ class STTfragment : Fragment() {
 
     override fun onPause() {
         super.onPause()
-        (activity as ChatActivity).fabsetting.show()
+	    if (activity is ChatActivity) {
+            (activity as ChatActivity).enableVoiceInput()
+            (activity as ChatActivity).fabsetting.show()
+        }
         recognizer.cancel()
         recognizer.destroy()
     }


kartik-karz:
https://github.com/fossasia/susi_android/issues/1713

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -285,6 +285,12 @@ class ChatActivity : AppCompatActivity(), IChatView {
             }
         }
 
+
+    }
```
arundhati24: What is being closed here?

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -341,6 +345,7 @@ class ChatActivity : AppCompatActivity(), IChatView {
             chatPresenter.check(true)
             btnSpeak.setImageResource(R.drawable.ic_mic_24dp)
             btnSpeak.setOnClickListener({
+                btnSpeak.isClickable = false
```
iamareebjamal: Should we disable the button instead?

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -122,6 +122,9 @@ class STTfragment : Fragment() {
 
     override fun onPause() {
         super.onPause()
+        if ( activity is ChatActivity) {
+            (activity as ChatActivity).enableVoiceInput()
+        }
         (activity as ChatActivity).fabsetting.show()
```
iamareebjamal: Move this inside the if block as well and remove the cast.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -122,6 +122,9 @@ class STTfragment : Fragment() {
 
     override fun onPause() {
         super.onPause()
+        if ( activity is ChatActivity) {
```
iamareebjamal: Extra space after `(`

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
kartik-karz: move the fabsetting.show()? shouldn't I do it in another PR? wouldn't it confuse others if they look at it later?

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
kartik-karz: both are kinda similar but disabled seems more intuitive, will change that.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
kartik-karz: also @iamareebjamal I think the cast is required by the compiler, it's an error if we remove the special cast.

iamareebjamal: I thought it should work through Kotlin smart cast. Let me check and get back to you

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
kartik-karz: the code was written between the last brace and so it felt as though there was an extra brace, it's resolved in the latest commit

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -122,8 +122,11 @@ class STTfragment : Fragment() {
 
     override fun onPause() {
         super.onPause()
-        (activity as ChatActivity).fabsetting.show()
+	    if (activity is ChatActivity) {
+            (activity as ChatActivity).enableVoiceInput()
+            (activity as ChatActivity).fabsetting.show()
```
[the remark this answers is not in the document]
kartik-karz: > I thought it should work through Kotlin smart cast. Let me check and get back to you

In our previous chat here https://github.com/fossasia/susi_android/pull/1714#discussion_r221815328
I explained the problem and you mentioned that you would test it out, was this after the test?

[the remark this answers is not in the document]
iamareebjamal: Yes it will work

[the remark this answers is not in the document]
kartik-karz: @iamareebjamal 
2018-10-12-200013_1366x768_scrot
I don't know why but I seem to have error compiling it right from that day, are you using some special option to make ide neglect that error or are you testing it directly using gradlew?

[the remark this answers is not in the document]
iamareebjamal: I checked general smart cast. It is not possible with activities, please continue with older fix

PR #1720, opened 2018-10-01
fix: Use correct color primary to make it consistent with other clients

### LINKED ISSUE #1719: Incorrect primary color used in the app
**Actual Behaviour**

The primary color used in SUSI.AI Android app is not the same as the color used in other clients.

**Expected Behaviour**

Update the primary color of the app to make it consistent with other SUSI clients.

Yes

arundhati24:
Fixes #1719 

Changes: Changed the color primary of the app to #4285f4 .

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="colorPrimary">#4184f3</color>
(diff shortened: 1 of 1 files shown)


iamareebjamal:
I think that there is not enough distinction in the primary and dark color. Maybe dark color should be darker or primary color should be lighter

arundhati24:
@iamareebjamal I agree. Actually, earlier both primary and dark color had the same hex code. I will check this out with the web client and will change the dark color accordingly.

PR #1781, opened 2018-11-21
fix: Improve UI of the login screen

### LINKED ISSUE #1671: Improve UI of the Login screen
**Actual Behaviour**

There are different types of fonts used for same context, spacing issues and so on, that need to be fixed.

**Expected Behaviour**

Improve the UI of the login screen and use the same color scheme as for the entire project.

Yes

aleronarjun:
Fixes #1671 

Changes: 

Made the Login/SignUp activities' UI similar to popular apps like facebook.
Added logic to Login/SignUp buttons for enable/disable
Made colors similar throughout
Corrected fonts
Vertically centered content
Made apt changes in landscape views
Made minor changes to margins and improved consistency of UI
Changed inFocus attribute so keyboard doesn't automatically popup

Screenshots for the change: 
screenshot_1542831684
screenshot_1542831687


aleronarjun:
@batbrain7 could you be more specific please? What do you mean by 'messenger' and why would strings here be consistent with it? Please clarify. Thanks.

PR #1811, opened 2018-12-14
Feature: Making Susi Default Step #1

zameelpichen:
Changes:

Initiated task to make susi default.

Issues, the voice does not work since, susi seems to depend on google assistant.

Note : I am making this pull request so that I can hear what other developers think about the issue, and to welcome your suggestion. This is my first attempt to contribute to susi. I appreciate any and all help.


zameelpichen:
@batbrain7 This is not solution to an issue, but a feature addition. What should I do? I have already written on top "Feature:". 

also, this is not a completed feature, but its initial functions are clean and completed. and can operate itself with the code in this pull request.

I would like others also join the building of this feature, which is why I am creating a pull request now.

This is a feature to to give susi android the ability to be used as the default assistant, i started doing this since i want to replace google assistant.

Please let me know how we should proceed.

iamareebjamal:
Please read this https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia/

An issue can be a bug, feature request. feature, documentation change, library upgrade, refactor, rant, or anything else.

Every PR **MUST** be linked to an issue

PR #1816, opened 2018-12-16
fix: GoogleAppIndexingWarning in AndroidManifest.xml (#1815)

### LINKED ISSUE #1815: Android Studio warning- App is not indexable by google search consider adding at least one activity with an action-view intent filter
**Actual Behaviour**

Android Studio shows warning message for AndroidManifest.xml file-
```
App is not indexable by Google Search; consider adding at least one Activity with an ACTION-VIEW intent-filler
```

**Expected Behaviour**

This is because app is missing deep-linking. 

**Steps to reproduce it**
Open AndroidManifest.xml file. If you have latest Android Studio (mine is 3.2), you will see this warning message.

**LogCat for the issue**

No logs.

**Screenshots of the issue**

susi-index-warning

Yes, definitely.

varunon9:
Fixes #1815

Changes: Added intent-filter in `WelcomeActivity` for google app-indexing. 
Open will open now when-
1. User tries to open a url: https://chat.susi.ai/app
2. Or programmatically via `susi.ai://` 

Screenshots for the change: 
susi-android-deep-linking

deep-link-susi

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -30,6 +30,14 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <!-- Accepts URIs that begin with "https://chat.susi.ai/app" -->
+                <data android:host="chat.susi.ai" android:pathPrefix="/app" android:scheme="https" />
+            </intent-filter>
         </activity>
 
         <activity


app/src/main/AndroidManifest.xml
```diff
@@ -30,6 +30,22 @@
...
+                <data android:host="chat.susi.ai" android:pathPrefix="/app" android:scheme="https" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <!-- Accepts URIs that begin with "susi.ai://" -->
+                <data android:scheme="susi.ai" />
+            </intent-filter>
```
iamareebjamal: This intent filter is too generic, we are not handling requests internally, so this will be useless.

varunon9: Any suggestion for URI? App should support deep linking, right?

iamareebjamal: Yes, but until it does not, we can't add this. So, remove this intent-filter.

iamareebjamal: 2nd should be removed. The problem in registering it before the app has deep linking support is that users will be confused and file issues regarding it. Even now users file multiple issues on single bug

varunon9: Updated PR

Issue #1819, opened 2018-12-21
App Crash : Change server preference is crashing the app


trancenoid:
**Actual Behaviour**
App is crashing when clicking on Change server backend in setting menu.

**Expected Behaviour**

App should not crash even if there is only one choice.

**Steps to reproduce it**

Goto to Setting and choose "Change server backend"

**LogCat for the issue**

```
2018-12-21 20:05:16.773 20625-20625/ai.susi E/AndroidRuntime: FATAL EXCEPTION: main
    Process: ai.susi, PID: 20625
    java.lang.IllegalStateException: preferences.getString(pr…, preferenceDefaultValue) must not be null
        at org.fossasia.susi.ai.helper.PrefManager.getString(PrefManager.kt:163)
        at org.fossasia.susi.ai.skills.settings.ChatSettingsFragment.showAlert(ChatSettingsFragment.kt:262)
        at org.fossasia.susi.ai.skills.settings.ChatSettingsFragment.access$showAlert(ChatSettingsFragment.kt:40)
        at org.fossasia.susi.ai.skills.settings.ChatSettingsFragment$onCreatePreferencesFix$7.onPreferenceClick(ChatSettingsFragment.kt:179)
        at android.support.v7.preference.Preference.performClick(Preference.java:1165)
        at android.support.v7.preference.Preference.performClick(Preference.java:1148)
        at android.support.v7.preference.Preference$1.onClick(Preference.java:172)
        at android.view.View.performClick(View.java:5612)
        at android.view.View$PerformClick.run(View.java:22285)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6123)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)
```

**Would you like to work on the issue?**
Yes

PR #1820, opened 2018-12-21
fix: App crash when trying to change server backend in settings

trancenoid:
Fixes : #1819 

Changes: Changed default value of CUSTOM_SERVER from null to "", which is required since the default value must be empty (not null) when the app is used for the first time.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
@@ -259,7 +259,7 @@ class ChatSettingsFragment : PreferenceFragmentCompat(), ISettingsView {
             flag = true
         }
         customerServer.isChecked = flag
-        inputUrlText.setText(PrefManager.getString(Constant.CUSTOM_SERVER, null))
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Please follow PR template

trancenoid:
@iamareebjamal Sorry about that. Fixed it.

app/src/main/java/org/fossasia/susi/ai/skills/settings/ChatSettingsFragment.kt
```diff
@@ -259,7 +259,7 @@ class ChatSettingsFragment : PreferenceFragmentCompat(), ISettingsView {
             flag = true
         }
         customerServer.isChecked = flag
-        inputUrlText.setText(PrefManager.getString(Constant.CUSTOM_SERVER, null))
+        inputUrlText.setText(PrefManager.getString(Constant.CUSTOM_SERVER, " "))
```
iamareebjamal: Use "" instead of " ", it is faster

trancenoid: Done!

PR #1824, opened 2018-12-28
	modified:   README.md 

Maimoi:
Fixes #[Add issue number here. Note: This will automatically close the issue. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Screenshots for the change:


iamareebjamal:
I think the change is unnecessary. This way, we can even add things like. Click on fork button and how to install git. This is beyond the scope of the project. Anyone wanting to contribute would at least know how to clone. We should only provide details specific to the project

PR #1869, opened 2019-01-22
fix: Static map open in chat on search location

### LINKED ISSUE #1793: Not show maps in chat
**Actual Behaviour**

In the chat screen, not show maps 
**Expected Behaviour** 
Should be show maps

**Steps to reproduce it**
1. Go to chat
2. Try typing where is singapore?

**LogCat for the issue**

**Screenshots of the issue**

screenshot_2018-12-06-07-41-39

NileshPranami:
Fixes #1793
 ->OpenStreetMap had stoped providing static map services
 ->Used token of mapBox for static map


app/src/main/java/org/fossasia/susi/ai/helper/MapHelper.kt
```diff
@@ -41,15 +42,16 @@ class MapHelper
...
+            mapURL = String.format(URL_SCHEME, longitude, latitude, zoom, width, height)
             webLink = String.format(MAP_URL_SCHEME, zoom, latitude, longitude)
             isParseSuccessful = true
         }
     }
 
     companion object {
-        private const val BASE_URL = "http://staticmap.openstreetmap.de/staticmap.php?"
-        private val URL_SCHEME = BASE_URL + "center=%f,%f&zoom=%f&size=%sx%s"
+        var MAPBOX_STATIC_MAP_KEY = BuildConfig.MAPBOX_API_KEY //create your own token by register on www.mapbox.com
```
iamareebjamal: Why is it var?

NileshPranami: @iamareebjamal Solved sir

app/build.gradle
```diff
@@ -12,6 +12,7 @@ apply plugin: 'kotlin-kapt'
 apply plugin: 'realm-android'
 
 def YOUTUBE_API_KEY = System.getenv('YOUTUBE_API_KEY') ?: "YOUR_API_KEY"
+def MAPBOX_API_KEY = '"'+System.getenv('MAPBOX_API_KEY')+'"' ?: '"DEFAULT"'
```
iamareebjamal: Extra single and double quotes

NileshPranami: Sir if we don't do so then it will not return string. In BuildConfig it will show key without string without double quotes.

iamareebjamal: I agree. But, then how is YOUTUBE_API_KEY working?

NileshPranami: I am not passing this key through menifest so. And this key is unable to fetch in menifest so I thought BuildConfig is better.

NileshPranami:
@iamareebjamal Thanks a lot sir

PR #1872, opened 2019-01-23
fix: Answer now contains the link which it should show #1758

### LINKED ISSUE #1758: No link to websearch result
**Actual Behaviour**

There is no link to visit (or anything) below text "Here is a websearch result:".

**Expected Behaviour**

There should be a link to visit or other field showing results.

**Steps to reproduce it**

\-

**LogCat for the issue**

none

**Screenshots of the issue**

screenshot_20181103-164907

I would like to help but I am afraid I cannot fix it on my own.

vaibhavg152:
Fixes #1758 [Add issue number here. Note: This will automatically close the issue. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Changes: Accessed all the links in the response returned by the api and added the first link to the answer. Previously the links were not being accessed in the correct way.

Link: https://api.susi.ai/susi/chat.json?country_code=&q=google%20tere%20naam&timezoneOffset=-330&latitude=28.54773708&country_name=&language=en&device_type=Android&longitude=77.18632999&geosource=gps 
the exact location of the links can be found out from the above link.

Screenshots for the change: 
Before:
screenshot_20190120-144228
After:
screenshot_20190123-145734


app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
```diff
@@ -42,9 +42,23 @@ class ParseSusiResponseHelper {
...
-                if (urlList.isEmpty()) isHavingLink = false
+
+                //get the Urls stored in 'data' of the answer object
+                val text = susiResponse.answers[0].data[0].get("object") //requires api warning suppressed
+                if (text != null) {
+                    val urlList = extractUrls(text)
+
+                    //appending the Urls to the answer
+                    //num_links is the maximum number of links to be appended
+                    val num_links = 1
```
batbrain7: use camel case only

iamareebjamal: No, wait for Valentine's Day! /s

app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
```diff
@@ -49,10 +49,10 @@ class ParseSusiResponseHelper {
                     val urlList = extractUrls(text)
 
                     //appending the Urls to the answer
-                    //num_links is the maximum number of links to be appended
-                    val num_links = 1
+                    //numLinks is the maximum number of links to be appended
+                    val numLinks = 1
                     for (l in urlList.indices) {
```
iamareebjamal: No single character variables

app/src/main/java/org/fossasia/susi/ai/chat/ParseSusiResponseHelper.kt
```diff
@@ -42,9 +42,23 @@ class ParseSusiResponseHelper {
...
+                if (text != null) {
+                    val urlList = extractUrls(text)
+
+                    //appending the Urls to the answer
+                    //numLinks is the maximum number of links to be appended
+                    val numLinks = 1
+                    for (link in urlList.indices) {
+                        if (link < numLinks)
+                            answer += "\n" + urlList[link]
+                    }
```
iamareebjamal: This is not C, directly iterate through link

Also, you don't even need to do that

```kotlin
isHavingLink = true
if (urlList.isNotEmpty()) {
    answer += "\n" + urlList[0]
} else {
    isHavingLink = false
}
```

vaibhavg152: This was done because I didn't know how many links should preferably to be included in the answer. Doing it this way leaves an option to easily change the number if required

iamareebjamal: What'll happen if we show all links?

vaibhavg152: For simple queries, there are a lot of links. Showing all the links will make the answer look messy and the user will have to scroll a lot to find the appropriate result.
Also the preview image for the link is also not displayed which makes the result even more ugly

iamareebjamal: OK, then we'll go with only 1 link

```kotlin
isHavingLink = true
if (urlList.isNotEmpty()) {
    answer += "\n" + urlList[0]
} else {
    isHavingLink = false
}
```

Issue #1807, opened 2018-12-10
App crashes when permissions are denied and the app is resumed after minimizing in Redmi 4A


khush3211:
**Actual Behaviour**
App crashes when permissions are denied and the app is resumed after minimizing in Redmi 4A. When partial or no permissions are given and app is minimized, on resuming a runtime null pointer exception occurs. Screenshot attached.

**Expected Behaviour**
No crash should happen.

**Steps to reproduce it**
Give partial or no permission in Android N on Redmi 4A (MiUI 10)

**LogCat for the issue**
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.google.android.packageinstaller/com.android.packageinstaller.permission.ui.GrantPermissionsActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2723)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2784)
	at android.app.ActivityThread.-wrap12(ActivityThread.java)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1523)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:163)
	at android.app.ActivityThread.main(ActivityThread.java:6238)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:794)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
	at com.android.packageinstaller.permission.ui.GrantPermissionsActivity.computePermissionGrantState(GrantPermissionsActivity.java:318)
	at com.android.packageinstaller.permission.ui.GrantPermissionsActivity.updateDefaultResults(GrantPermissionsActivity.java:367)
	at com.android.packageinstaller.permission.ui.GrantPermissionsActivity.onCreate(GrantPermissionsActivity.java:124)
	at android.app.Activity.performCreate(Activity.java:6868)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1119)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2676)
[...]

PR #1875, opened 2019-01-23
fix: removed null values from a string array before passing to the function. #1807

vaibhavg152:


immadisairaj:
@vaibhavg152 Every PR should be associated with some issue and should have a description based on the PR template

PR #1891, opened 2019-01-29
fix: Replaced single letter variable names with proper names.

### LINKED ISSUE #1688: Remove single character variable name
**Actual Behaviour**

There are many single variables name used in project

**Expected Behaviour**

There should be proper declaration of them with proper naming.

Yes!

yashk2000:
Fixes #1688 

Changes: Replaced all single character variable names with proper names.


iamareebjamal:
Thanks a lot :smile:

PR #1725, opened 2018-10-01
feat: Add login option in the overflow menu

Princejhabak:
Fixes #1703

Changes: Login option added to the overflow menu which will take the user to the login screen

Screenshots for the change:
46090175-c7d6ce00-c1cd-11e8-9aac-11fdfb8fcb0c

PR #1929, opened 2019-02-10
feat: Added Login Logout option in menu bar

### LINKED ISSUE #1703: Add signin / signup option in the menu
**Actual Behaviour**

User is not able to sign in / sign up later if he skips the sign in /sign up process when the app starts for the first time

**Expected Behaviour**

User must be able to sign in / sign up anytime he wants

**Steps to reproduce it**

Option for sign in / sign up can be added to  the menu  which will take the user to sign in page

Yes , I would like to work on this issue

NileshPranami:
Fixes #1703
and resolve PR #1725


immadisairaj:
@iamareebjamal @batbrain7 
I think we can close this PR as this is already fixed in the PR #2046

PR #1972, opened 2019-02-20
chore: updated settings activity image in docs

naman653:
Changes: Settings activity layout has been changed but in docs folder it contains the old image of settings activity.


naman653:
You can view the file in commit section, as I have deleted the old file and replaced it with a new file, it is just showing the decrease in image size in view commit page. If you view the file, you can see the updated image.

PR #1973, opened 2019-02-21
chore(strings): updated error messages in multiple languages

naman653:
Changes: If the internet is off, skills activity is not showing that the reason for being unable to fetch skills can be due to the internet connection. It just shows that an error has occurred. So I've updated the message to point user to check internet connection.


iamareebjamal:
It was not a feat, it was a chore

PR #1976, opened 2019-02-22
fix: pull request template was saying 'issue' instead of 'commit'

naman653:
Changes: Now PR template is saying "commit" instead of "issue"


iamareebjamal:
It's correct in what it says

PR #1983, opened 2019-02-25
fix: remove hungarian notation

### LINKED ISSUE #1689: Remove hungarian notation from project
**Actual Behaviour**

There are Hungarian notation used in the project for naming the variables

**Expected Behaviour**

They should not be used.

You can take it if you can

immadisairaj:
Fixes #1689, #1688

Changes: Removed Hungarian notations in all the files, Replaced single variable names with appropriate names, changed shortcut variable names with appropriate names


app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -287,11 +287,11 @@ class ChatActivity : AppCompatActivity(), IChatView {
                     }
                 })
 
-                val ttsParams = HashMap<String, String>()
```
iamareebjamal: Revert

immadisairaj:
@iamareebjamal made changes as requested. Can you please review this.

PR #1991, opened 2019-03-02
fix: User's message should be ticked before SUSI's reply

### LINKED ISSUE #1990: User's message should be ticked before SUSI's reply

Right now, the user's message is getting a tick mark after the reply of SUSI. This doesn't seem natural since the reply should come after the message is sent.

**Screenshots**

susi-tick

Yes

balaji-ramavathu:
Fixes #1990 

Changes: The check to display tick is changed in a way that if the network connection is available, the message gets ticked indicating that the message is sent.  

Screenshots for the change: 

susi-tick-pr


balaji-ramavathu:
@iamareebjamal please review :-)

balaji-ramavathu:
Any updates here?

iamareebjamal:
We can't show a tick unless we are sure that the message has reached to the server

arundhati24:
As rightly mentioned by @iamareebjamal , we cannot show a tick unless we are sure that the message has reached to the server. If this has to be done, maybe a slight delay could be introduced after showing a tick (when the message has been delivered) and before displaying the SUSI response, but I am not sure whether this needs to be done for now. @iamareebjamal It would be very helpful if you could please reflect upon this a bit so that we can finalise this as well?

PR #1993, opened 2019-03-03
fix: fixes crash

### LINKED ISSUE #1992: App crashes due IllegalStateException
**Describe the bug**

The swipe_refresh_layout becomes null when the fragment is changed from SkillListing and crashes

**Expected Behaviour**

The app shouldn't crash in that case

**Steps to reproduce it**

1. Open the application
2. Click the Susi logo
3. Immediately go into settings page or about us page
4. wait for some time
5. app will crash

**LogCat for the issue**

2019-03-03 12:04:31.040 3403-3403/ai.susi E/AndroidRuntime: FATAL EXCEPTION: main
    Process: ai.susi, PID: 3403
    java.lang.IllegalStateException: swipe_refresh_layout must not be null
        at org.fossasia.susi.ai.skills.skilllisting.SkillListingFragment.updateSkillsAdapter(SkillListingFragment.kt:99)
        at org.fossasia.susi.ai.skills.skilllisting.SkillListingPresenter.onSkillFetchSuccess(SkillListingPresenter.kt:86)
        at org.fossasia.susi.ai.data.SkillListingModel$fetchSkills$1.onResponse(SkillListingModel.kt:48)
        at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
        at android.os.Handler.handleCallback(Handler.java:836)
        at android.os.Handler.dispatchMessage(Handler.java:103)
        at android.os.Looper.loop(Looper.java:203)
        at android.app.ActivityThread.main(ActivityThread.java:6364)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1076)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:937)

Yes

immadisairaj:
Fixes #1992

Changes:
Whenever the cancel is called the fetching process is continuing. So, if we put it in the while loop till all the calls are cancelled then it'll be working fine.

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/skills/skilllisting/SkillListingPresenter.kt
@@ -51,6 +51,8 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
     }
 
     override fun onGroupFetchSuccess(response: Response<ListGroupsResponse>) {
+        if (skillListingView == null)
+            return
         val groupsResponse = response.body()
         if (response.isSuccessful && groupsResponse != null) {
             Timber.d("GROUPS FETCHED")
@@ -71,6 +73,8 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
     }
 
     override fun onSkillFetchSuccess(response: Response<ListSkillsResponse>, group: String) {
+        if (skillListingView == null)
+            return
         val skillsResponse = response.body()
         skillListingView?.visibilityProgressBar(false)
         if (response.isSuccessful && skillsResponse != null) {
@@ -98,6 +102,8 @@ class SkillListingPresenter(val skillListingFragment: SkillListingFragment) : IS
     }
 
     override fun onSkillMetricsFetchSuccess(response: Response<ListSkillMetricsResponse>) {
+        if (skillListingView == null)
+            return
         val skillsMetricsResponse = response.body()
         skillListingView?.visibilityProgressBar(false)
         if (response.isSuccessful && skillsMetricsResponse != null) {


app/src/main/java/org/fossasia/susi/ai/data/SkillListingModel.kt
```diff
@@ -73,9 +73,12 @@ class SkillListingModel : ISkillListingModel {
...
         try {
-            authResponseCallGroups.cancel()
-            authResponseCallSkills.cancel()
-            authResponseCallMetrics.cancel()
+            while (!authResponseCallGroups.isCanceled)
+                authResponseCallGroups.cancel()
+            while (!authResponseCallSkills.isCanceled)
+                authResponseCallSkills.cancel()
+            while (!authResponseCallMetrics.isCanceled)
+                authResponseCallMetrics.cancel()
```
iamareebjamal: This is strangest use of while loop I have ever seen!

immadisairaj: Can you please suggest what can be done instead

iamareebjamal: There is no need of while loop. Simply  solve the memory leak using common practices

immadisairaj: Can I use Handler().postDelayed

iamareebjamal: No

immadisairaj: Can I use boolean to check if the fetch is called or not

iamareebjamal: You simply need to return from the callbacks if the presenter is destroyed

immadisairaj:
@iamareebjamal 
I checked if the presenter is destroyed then it calls cancelFetch(). I just returned from not fetching before it starts to fetch. But if cancelFetch() is called after the fetching starts then I cancel all the calls

iamareebjamal:
I'm sorry but you're not able to understand what I'm trying to say. Analyse the situation in more detail and tell me why null pointer exception is happening and why not just returning before the call from where NPE is happening won't solve the problem

immadisairaj:
@iamareebjamal 
This is what I understood. In the SkillListingPresenter we are changing the skillListingView to null. All the fragment views will become null. When we are fetching skills we are also updating the views in the fragment where there are no views. So, Checked whether the view is null initially and returned without fetching anything.
Is this what you were saying or did I again misunderstood?

PR #2011, opened 2019-03-11
feat: Move from signup page to login page

### LINKED ISSUE #2009: Feature: Option to redirect to Login page for existing user from Sign Up page
There should be an option in the Sign Up page, clicking on which will redirect to Login Page.

**Additional context**
This feature targets those user who already have account in Susi but tries to register again mistakenly. This will allow them to redirect to Login page, and hence they can login from here.

**Screenshots**
login

Yes

atm1504:
Fixes #2009 

Changes: 
* LoginActivity.kt
* SignUpActivity.kt
* activity_sign_up.xml
* strings.xml

Screenshots for the change: 
signUpToLogin

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -156,6 +156,7 @@
     <string name="share_items">Share</string>
 
     <string name="sign_up">Sign Up</string>
+    <string name="signUpToLogin">Existing User?\n Click here to Login</string>
 
     <string name="url_cannot_be_empty">URL cannot be empty</string>
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -156,7 +156,7 @@
     <string name="share_items">Share</string>
 
     <string name="sign_up">Sign Up</string>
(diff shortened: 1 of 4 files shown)


Bhavnaharitsa:
Is n't the app crashing when entering into settings 
Just asking cause it is crashing for me ? :)

atm1504:
> Is n't the app crashing when entering into settings
> Just asking cause it is crashing for me ? :)

No, it is working fine in my mobile

atm1504:
@iamareebjamal @batbrain7 @arundhati24  please review it

app/src/main/res/values/strings.xml
```diff
@@ -156,6 +156,7 @@
     <string name="share_items">Share</string>
 
     <string name="sign_up">Sign Up</string>
+    <string name="signUpToLogin">Existing User?\n Click here to Login</string>
```
iamareebjamal: Strange name. Should be `existing_user_login`

See that every other string is snake case and you have used camel case

atm1504: I will correct my mistake sir soon

atm1504:
@iamareebjamal I have made the requested changes. Please review it now

PR #2044, opened 2019-03-20
feat: Added skip button in signup page

### LINKED ISSUE #2041: Add skip button in SignUp page
Add Skip button in signup page. The work of the button will be similar to that of the skip button present in login page.

**Additional context**
If some users open up the signup page by mistake, but he wants to skip it, then he could press this button to skip the signup page.

**Screenshots**
signup
login

Yes

atm1504:
Fixes #2041

Changes: 
* Added skip option in activity_sign_up.xml
* Added the intent in SignUpActivity
* Added the strings in string files

Screenshots for the change: 
skipSignUpGif

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/signup/SignUpActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/signup/SignUpActivity.kt
@@ -13,6 +13,7 @@ import android.view.MenuItem
 import android.view.View
 import kotlinx.android.synthetic.main.activity_sign_up.*
 import org.fossasia.susi.ai.R
+import org.fossasia.susi.ai.chat.ChatActivity
 import org.fossasia.susi.ai.login.LoginActivity
 import org.fossasia.susi.ai.helper.AlertboxHelper
 import org.fossasia.susi.ai.helper.Constant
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/fossasia/susi/ai/signup/SignUpActivity.kt
```diff
@@ -84,6 +85,16 @@ class SignUpActivity : AppCompatActivity(), ISignUpView {
         signUpToLoginPage()
         cancelSignUp()
         signUpToTermsConditionPage()
+        skipSignUp()
+    }
+
+    fun skipSignUp() {
+        skipSignUp.setOnClickListener {
+            val intent = Intent(this@SignUpActivity, ChatActivity::class.java)
+            intent.putExtra(Constant.FIRST_TIME, false)
```
iamareebjamal: Why is this false?

app/src/main/res/values/strings.xml
```diff
@@ -173,6 +173,7 @@
     <string name="activity_login_log_in">Log in</string>
     <string name="activity_login_sign_up">Sign up for SUSI</string>
     <string name="activity_login_skip">Skip</string>
+    <string name="activity_signup_skip">Skip</string>
```
iamareebjamal: What's the difference between this and the string above?

atm1504: Sir, I added the string only because those two strings belong to two different pages. So, I thought it would be better to seperate to define it sepeartely. If asked i will remove it.

iamareebjamal: No, strings don't depend on their page, only on their content

PR #2064, opened 2019-03-22
fix: Disable signup when logged in

### LINKED ISSUE #2063: "Signup" in menu bar is clickable even when user is logged in
**Describe the bug**
Signup button in menu is clickable even when the user has already logged in.

**Expected Behaviour**
It should be disabled when the user has logged in. 

**Steps to reproduce it**
Add setEnabled false in the kotlin file when user is logged in.

**Screenshots of the issue**
signupenable

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Redmi Note 5 Pro |
|Android Version| Oreo 8.1   |

Yes

atm1504:
Fixes #2063 

Changes:
Added "Sign Up" option property setEnabled as false when user is logged in.

Screenshots for the change: 
signupdisable


iamareebjamal:
Now please focus on real issues instead of adding menus and login logout

PR #2072, opened 2019-03-25
fix: Placeholder values of author name and example is not showing now in GroupWise Skills

### LINKED ISSUE #2071: GroupWise skills showing placeholder values of example and author name
**Describe the bug**
GroupWise skills fragment is showing placeholder values for author name and example, i.e. "<author_name>" for an author name and "\<The question that should be shown in public skill displays\>" for an example.

**Expected Behaviour**
Author name should be set to --> "Author not found"
The example should be set to --> ""

**Steps to reproduce it**
Click on any category of skills and try to search for skills with author name as <author_name> and example as "\<The question that should be shown in public skill displays\>"

Yes

naman653:
Fixes #2071 

Changes: Now placeholder values of example and author name of skill data is not showing

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -438,10 +438,6 @@
     <string name="report_send_success">Skill reported successfully</string>
     <string name="reported_already">Skill already reported</string>
 
-    <!--Default placeholder values of skills description-->
-    <string name="default_author_name" translatable="false">&lt;author_name&gt;</string>
-    <string name="default_example" translatable="false">&lt;The question that should be shown in public skill displays&gt;</string>
-
     <!--Privacy -->
     <string name="privacy_welcome">Welcome to SUSI.AI</string>
     <string name="privacy_welcome_details">Thanks for using our products and services (“Services”). The Services are provided by SUSI Inc. (“SUSI”), located at 93 Mau Than, Can Tho City, Viet Nam. By using our Services, you are agreeing to these terms. Please read them carefully.</string>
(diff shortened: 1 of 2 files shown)


naman653:
It was just checking if the author name or example is empty or not. But if there was default placeholder value present, it sets these values to those default ones showing author name as <author_name> and same as for example.

> Why was it showing the placeholder values?

iamareebjamal:
But from where were the placeholder values coming from?

naman653:
These values are directly coming from the server along with skills data. There is a template to create a new skill, and if the user doesn't replace placeholder values while creating skills, then these values are set in as data.

app/src/main/res/values/strings.xml
```diff
@@ -438,6 +438,10 @@
     <string name="report_send_success">Skill reported successfully</string>
     <string name="reported_already">Skill already reported</string>
 
+    <!--Default placeholder values of skills description-->
+    <string name="default_author_name" translatable="false">&lt;author_name&gt;</string>
+    <string name="default_example" translatable="false">&lt;The question that should be shown in public skill displays&gt;</string>
```
iamareebjamal: Remove from strings and make String constants

naman653: I thought of that, but the same problem exists in the skill details fragment. So I thought these strings might be useful in the future. Please enlighten, if asked changes should be done or not.

iamareebjamal: They should be done

PR #2093, opened 2019-03-28
feat: Speaks welcome message when voice used for first time

### LINKED ISSUE #2087: Susi should speak up 'How can I help you? ' when susi voice is used for first time.
We can see in Google Assistant, Google speaks up 'How he can help?'  when users open it for first time. We should have a similar feature here, so that it creates a lively nature. 

**Additional context**
It is necessary so that the user gets to know that susi voice is live. 

Yes

atm1504:
Fixes #2087

Changes: 
* Added tts in STTfragment
* Added few strings
* Added preference value in Welcome.kt

Screenshots for the change: 
ezgif com-video-to-gif


atm1504:

atm1504:
@batbrain7 please review it

atm1504:

atm1504:
@iamareebjamal @batbrain7 @arundhati24 please review it.

atm1504:
@iamareebjamal @batbrain7 @arundhati24 please review it.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -39,17 +44,42 @@ class STTFragment : Fragment() {
...
             thisActivity.fabsetting.hide()
-        promptSpeechInput()
+
+        textToSpeech = TextToSpeech(context, this)
+        val used_voice_state = PrefManager.getBoolean(R.string.used_voice, false)
+        if (used_voice_state == false) {
+            Handler().postDelayed({
+                speakOut()
+                PrefManager.putBoolean(R.string.used_voice, true)
+            }, 500)
```
iamareebjamal: Why

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -39,17 +44,42 @@ class STTFragment : Fragment() {
...
+            Handler().postDelayed({
+                promptSpeechInput()
+            }, 1500)
```
iamareebjamal: Why

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
atm1504: this is because we cannot start the speech engine as soon as the fragment on. The speech engine was failing to start. Upon, google search, I found that we should delay it by few moments to start the engine. After doing so, I found it worked.

iamareebjamal: This is creating a memory leak

arundhati24: I agree with what @iamareebjamal said. This is creating a memory leak because the anonymous runnable holds an implicit reference to your activity. Now, if your activity is destroyed within the delay duration after its creation, it would lead to a memory leak because the runnable will still hold a reference to the activity which will prevent the activity from being garbage collected.
Moreover, this code should not be put inside the fragment.

atm1504: @arundhati24 @iamareebjamal I will try to find a better solution and update the pr.

atm1504:
@iamareebjamal @arundhati24 please review it now. This time there was no memory leak. Also, I have used the postDelay function, in this case, to call the speechPrompt input after the message is spoken by complete. The message is spoken only once when the user uses voice input for the first time, without clearing the app from background.

atm1504:
@iamareebjamal @arundhati24 @batbrain7 please review it.

atm1504:
@iamareebjamal @arundhati24 please review it.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -39,17 +43,37 @@ class STTFragment : Fragment() {
...
+                        PrefManager.putBoolean(R.string.used_voice, true)
+                        Handler().postDelayed({
+                            promptSpeechInput()
+                        }, 1500) // Starting speech engine after welcome message is spoken
+                    } else {
+                        promptSpeechInput()
+                    }
+                }
+            })
+        }
```
arundhati24: All this logic should be placed in the presenter. The view should only know what to show to the user.

atm1504: There is no presenter class made for STT Fragment. Also, I found that the logic of the other functions was also written in the fragment. Hence, I wrote the code in the fragment itself.

mariobehling:
Circle Ci is failing. Please fix.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
atm1504: @arundhati24 what should I do? Shall I make presenter class for the fragment or just keep it here>

iamareebjamal: Clear all references to Handler onDestroyView

atm1504:
Apk file: 
[welcome_message.apk.zip](https://github.com/fossasia/susi_android/files/3280776/welcome_message.apk.zip)

atm1504:

atm1504:
@iamareebjamal @arundhati24 please review it.

atm1504:
@iamareebjamal @arundhati24 @batbrain7 please review it

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -40,17 +46,44 @@ class STTFragment : Fragment() {
...
         if (thisActivity is ChatActivity)
             thisActivity.fabsetting.hide()
-        promptSpeechInput()
         setupCommands(rootView)
         return rootView
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        runnable = Runnable {
+            Handler().post {
```
iamareebjamal: Not removed

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -40,17 +46,44 @@ class STTFragment : Fragment() {
...
+                textToSpeech = TextToSpeech(requireContext(), TextToSpeech.OnInitListener { status ->
+                    if (status != TextToSpeech.ERROR) {
+                        val locale = textToSpeech?.language
+                        textToSpeech?.language = locale
+                        if (!PrefManager.getBoolean(R.string.used_voice, false)) {
+                            textToSpeech?.speak(getString(R.string.voice_welcome), TextToSpeech.QUEUE_FLUSH, null)
+                            PrefManager.putBoolean(R.string.used_voice, true)
+                            delayRunnable = Runnable {
+                                Handler().postDelayed({
```
iamareebjamal: Not removed

atm1504:
I have removed the runnable's inside  onDestroyView()

iamareebjamal:
Those runnables are not even registered. Your operation is no-op

atm1504:
@iamareebjamal please review now. I think the latest commit is the thing that you wanted.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -187,7 +184,7 @@ class STTFragment : Fragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
-        Handler().removeCallbacks(runnable)
-        Handler().removeCallbacks(delayRunnable)
+        mainHandler.removeCallbacksAndMessages(null)
+        subHandler.removeCallbacksAndMessages(null)
```
iamareebjamal: Useless. You aren't passing the runnable to cancel

atm1504:
@iamareebjamal this time I created a reference for the handler and runnable. I have used the handler reference to clear the runnable's. Please review now.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -29,6 +32,11 @@ class STTFragment : Fragment() {
     lateinit var recognizer: SpeechRecognizer
     lateinit var chatPresenter: IChatPresenter
     private val thisActivity = activity
+    private var textToSpeech: TextToSpeech? = null
+    private lateinit var runnable: Runnable
+    private lateinit var delayRunnable: Runnable
+    var mainHandler: Handler = Handler()
+    var subHandler: Handler = Handler()
```
iamareebjamal: Where are you changing this variable?

iamareebjamal: And from which external class are you accessing these variables?

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -145,7 +180,16 @@ class STTFragment : Fragment() {
...
+            textToSpeech?.stop()
+        }
         recognizer.cancel()
         recognizer.destroy()
     }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        mainHandler.removeCallbacks(runnable)
+        subHandler.removeCallbacks(delayRunnable)
```
iamareebjamal: Useless as the runnable registered in the handler is not removed here

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
atm1504: > And from which external class are you accessing these variables?

This is prompting up because I reset the head 1 commit behind. The variables of handlers were already defined there

atm1504:
> Still wrong

I think what you mean is that, we should have the contents of the runnables written when we are declaring it in global scope. Now, we havt to use that to run it in ViewCreate, and destroy that on DestroyView. Initially, what we had is that, the usages of the variables showed that the thing that I destroyed was not the one I wanted. Bit, now, I  hecked that the references are okk. And its working fine. Please review the last commit.

app/src/main/java/org/fossasia/susi/ai/chat/STTfragment.kt
```diff
@@ -29,6 +32,9 @@ class STTFragment : Fragment() {
     lateinit var recognizer: SpeechRecognizer
     lateinit var chatPresenter: IChatPresenter
     private val thisActivity = activity
+    private var textToSpeech: TextToSpeech? = null
+    private var mainHandler: Handler = Handler()
+    private var subHandler: Handler = Handler()
```
iamareebjamal: Where are you changing these variables?

atm1504:
@iamareebjamal I did read about it from https://www.quora.com/How-do-I-stop-a-running-handler-in-Android and updated the pr in the way mentioned here. Please review it.

atm1504:
@iamareebjamal please review it now

atm1504:

PR #2060, opened 2019-03-22
fix: Add multilanguage support

atm1504:
Fixes #2001 

Changes:
* Added the setLocate function in ChatSettingsFragment.kt. 
This function is changing the language of the app

Screenshots for the change: 
multiLang

PR #2117, opened 2019-03-31
chore: intent to skills activity disabled on language change

naman653:
**Changes:**
On changing the language in settings, it directly starts an intent to the skills activity. But this should not be the case as while changing settings, a user would never want to be migrated to somewhere else, especially while changing a language.

**P.S**.-> These changes are just based on the reviews from my friends. These are not much relevant.


atm1504:
Is the translation working when you are removing the intent. As per I know we need to restart the activity for changing the language of the app.

naman653:
This language change is meant for changing the language of skills, not for changing the basic language.

atm1504:
This function is extending the multi language support. It won't work if we don't restart the activity.
Multi language is added in #2060.

atm1504:
@iamareebjamal I hope we should not have this. If it happens then multi-language support  won't work.

iamareebjamal:
@naman653 is correct

PR #2121, opened 2019-03-31
Added toast messages on responding

Bhavnaharitsa:
Fixes #2120 

Changes: Added toast message 

Screenshots for the change: 
Screenshot_20190331-183429_SUSIAI

PR #2125, opened 2019-04-01
Toast added when language is changed

Bhavnaharitsa:
Changes: Whenever a user changes languages it simply directs to the Skills page . Instead, he should know that the language is changed by a simple toast to which language it is changed.

Screenshots for the change: 
Screenshot_20190401-144842_SUSIAI


Bhavnaharitsa:
@iamareebjamal please review this !!

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
```diff
@@ -104,6 +104,7 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
                                 thumbsUp?.setImageResource(R.drawable.thumbs_up_solid)
                                 model.skillLocation?.let { location -> rateSusiSkill(Constant.POSITIVE, location, context) }
                                 setRating(true, true)
+                                Toast.makeText(context,"Thank you for responding", Toast.LENGTH_SHORT).show()
```
arundhati24: Is this change for this PR?

Bhavnaharitsa: Yes it is! Because I thought the user might know if the language has changed if there's no toast message ,like an intimation. So I thought it is important

arundhati24: I can see same changes in #2121 . Why so?

Bhavnaharitsa: That is probably because I couldn't cherry pick and squash commits. I shall see that other pull request commits don't enter here.
May I ask you, Can closing the other pull request solve the issue for now? Because this one has the commit of other pull request as well.

PR #2139, opened 2019-04-02
Added link of Issues on Help page 

Bhavnaharitsa:
Changes: In the help page, the steps are shown on how to report issues but I think providing URL of the Issue page would be helpful and hence attached the URL of the help page . A user can now access the help page from both About Us and Help page.

Screenshots for the change: 
Screenshot_20190402-200943_SUSIAI


atm1504:
@Bhavnaharitsa are you getting directed to the webpage by clicking on 'Report Issues'. I hope just by adding link it won't go. If it is a case that you are not redirected to the webpage, then what is the need of adding the link. 
I am not sure that it is directing or not. Please do check it.

PR #2140, opened 2019-04-03
Login page on back pressed goes to Skills Page 

### LINKED ISSUE #2136: Login page from menu bar closes the app after pressing back
**Describe the bug**
If the user hasn't logged in, he clicks the login page under menu in SkillsActivity and it directs to the login page . But for some reasons if he doesn't want to login and clicks back button,the app closes down. 

**Expected Behaviour**
On clicking back button, it should go back to menu bar in SkillsAcitivity so that s/he can continue using the app.
The same can be implemented for signup page as well.

Yes

Bhavnaharitsa:
Fixes #2136 
Changes: Made changes in LoginActivity and AndroidManifest 

Screenshots for the change: 

ezgif com-video-to-gif

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -50,22 +50,29 @@
         <activity
             android:name=".skills.SkillsActivity"
             android:configChanges="orientation|screenSize"
-            android:theme="@style/PreferencesThemeLight">
+            android:theme="@style/PreferencesThemeLight"
+            >
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value="org.fossasia.susi.ai.chat.ChatActivity" />
         </activity>
 
         <activity
-            android:name=".signup.SignUpActivity"
+             android:name=".signup.SignUpActivity"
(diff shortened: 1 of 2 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -50,22 +50,29 @@
...
-            android:name=".signup.SignUpActivity"
+             android:name=".signup.SignUpActivity"
             android:theme="@style/SignUp_Theme"
             android:windowSoftInputMode="stateVisible|adjustResize"
-            tools:replace="android:theme" />
+            tools:replace="android:theme"
+            >
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".skills.SkillsActivity" />
```
iamareebjamal: Completely wrong solution and we don't need these enhancements

Bhavnaharitsa: Sir,should I remove revert the changes in Manifest file ?

iamareebjamal: Yes

PR #2151, opened 2019-04-09
Ripple effect

Bhavnaharitsa:
Changes: The list of words though they are clickable does not indicate that they are being clicked.
inorder to indicate the user that the option has been clicked, I added the ripple effect.


iamareebjamal:
Various unrelated changes in PR

Bhavnaharitsa:
This might be the case of executing various tasks, shall correct them! @iamareebjamal

PR #2167, opened 2019-04-13
feat: Implement tap targets for butttons in chat activity

### LINKED ISSUE #2166: Implement tap targets for chat activity buttons
Currently user is not able to judge the use of the floating action buttons in the chat activity.Therefore we can use tap targets for these buttons to help user.These tap targets will only be shown at first installation of the app. 

Yes

chanmol1999:
Fixes #2166

Changes:chat_activity,strings.xml
Screenshots for the change: 
video_20190413_150542

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -22,7 +22,6 @@ import android.os.Handler
 import android.os.Message
 import android.speech.tts.TextToSpeech
 import android.speech.tts.UtteranceProgressListener
-import android.support.annotation.RequiresApi
 import android.support.design.widget.Snackbar
 import android.support.v4.app.ActivityCompat
 import android.support.v4.view.GestureDetectorCompat
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -5,6 +5,7 @@ import ai.kitt.snowboy.audio.AudioDataSaver
 import ai.kitt.snowboy.audio.RecordingThread
 import android.Manifest
 import android.annotation.SuppressLint
+import android.annotation.TargetApi
 import android.app.ProgressDialog
 import android.content.BroadcastReceiver
 import android.content.Context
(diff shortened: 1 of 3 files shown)


chanmol1999:
Please review @iamareebjamal @batbrain7 @arundhati24

chanmol1999:
Please review sir @iamareebjamal @batbrain7 
Do I need to make any changes.

iamareebjamal:
Peer Review

Bhavnaharitsa:
please add the translations in all the languages for the new strings you have added.

Bhavnaharitsa:
But, do we need this feature in the first place? Because I don't think most of the android chat apps have this feature.
@chanmol1999 This was my opinion. If you feel this is a good feature, then you can keep it and let the mentors decide if we need this or not.

Bhavnaharitsa:
@iamareebjamal I have reviewed
Thank you.

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -82,7 +88,6 @@ class ChatActivity : AppCompatActivity(), IChatView {
             textToSpeech?.stop()
         }
     }
-
```
immadisairaj: Why is this line removed?

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -111,6 +116,50 @@ class ChatActivity : AppCompatActivity(), IChatView {
         }
     }
 
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
```
immadisairaj: So the minimum version needed to use this feature is Lollipop.

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -111,6 +116,50 @@ class ChatActivity : AppCompatActivity(), IChatView {
...
+                                        .transparentTarget(false)
+                                        .icon(getDrawable(R.drawable.ic_open_search))// Specify whether the target is transparent (displays the content underneath) // Specify a custom drawable to draw as the target
+                                        .targetRadius(100)),
+                        TapTargetView.showFor(this, // `this` is an Activity
+                                TapTarget.forView(findViewById(R.id.fabsetting), getString(R.string.susi_icon_title), getString(R.string.susi_icon_description))
+                                        // All options below are optional
+                                        .outerCircleColor(R.color.colorPrimaryDark)
+                                        .outerCircleAlpha(0.96f)
+                                        .targetCircleColor(R.color.md_lime_50)
+                                        .titleTextSize(30)
```
immadisairaj: Don't use hard coded dimensions

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
Bhavnaharitsa: but we don't use lollipop

PR #2197, opened 2019-05-11
feat: Toast appears when user rates chat

### LINKED ISSUE #2120: Toast message after giving response
In the chat activity , after the user gives response ( thumbs up or thumbs down ) it just goes off bluntly and the user does not always rate the skill

**Additional context**
A toast message thanking the user for responding would be better so that it gives a better experience.

**Screenshots**
Screenshot_20190331-182346_SUSIAI

Yes

atm1504:
Fixes #2120 

Screenshots for the change: 
ezgif com-video-to-gif

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
@@ -39,6 +39,7 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
     val thumbsDown: ImageView? by bindOptionalView(R.id.thumbs_down)
 
     private var model: ChatMessage? = null
+    private lateinit var contexts: Context
 
     /**
      * Inflate ChatView
@@ -49,6 +50,7 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
     fun setView(model: ChatMessage?, viewType: Int, context: Context) {
         if (model != null) {
             this.model = model
+            contexts = context
             try {
                 when (viewType) {
                     ChatFeedRecyclerAdapter.USER_MESSAGE -> {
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
```diff
@@ -130,8 +130,10 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
         realm.beginTransaction()
         if (which) {
             model?.isPositiveRated = what
+            Toast.makeText(contexts, R.string.rate_chat, Toast.LENGTH_SHORT).show()
         } else {
             model?.isNegativeRated = what
+            Toast.makeText(contexts, R.string.rate_chat, Toast.LENGTH_SHORT).show()
```
iamareebjamal:
```js
if (i ==1)
  print('Yes')
else
  print('Yes')
```

What's the difference?

atm1504: Sir, I am making it in one line. But, here look that it has been succesfully added to the database.

app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
```diff
@@ -49,6 +50,7 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
     fun setView(model: ChatMessage?, viewType: Int, context: Context) {
         if (model != null) {
             this.model = model
+            contexts = context
```
iamareebjamal: Why?

atm1504: I tried to get the context via requireContext in setRatings, but I didn't get it. Also getContext, this failed. Hence, I came up with this id of storing the context in a global variable and using it in this function.

iamareebjamal: No need to store in global variable, just pass it as an argument to the function

atm1504: > Why?

Okk sir

atm1504:
@iamareebjamal sir, I further analysed it and found that the method I implemented is wrong, I will further look into it and make relevant changes.

PR #2198, opened 2019-05-12
feat: Swap between login and signup page

### LINKED ISSUE #2097: Swipe to Interchange between Login and SignUp Activity.
* Swipe from right to left to open SignUp page from Login Page.
* Swipe from Left to Right to open Login Page from Signup page

**Additional context**
Now a days, apps come with bottom navigation. In most of the apps, we can open signup from login page or login from signup page just by swapping fingers over the screen. I think we should have similar feature in our app.

Yes

atm1504:
Fixes #2097 

Changes: 
Id gave to the scrollview of login and signup activity
Using this id, onTouch Listener has been implemented.
Now user can move from login to signup page and vice versa by swap gesture

Screenshots for the change: 
ezgif com-video-to-gif


iamareebjamal:
Not needed

PR #2201, opened 2019-05-14
fix: Made search, susi icon invisible in stt fragment

### LINKED ISSUE #2177: chat search icon appears in Stt fragment
**Describe the bug**
The chat search icon shows up in stt fragment as well. It is only required in chatActivity and not needed in stt fragment.
We need to remove it

Screenshots:
Screenshot_20190417-063220_SUSIAI

Yes

atm1504:
Fixes #2177 

Screenshots for the change: 
ezgif com-video-to-gif

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
@@ -109,6 +109,10 @@ class ChatActivity : AppCompatActivity(), IChatView {
                 chatPresenter.startComputingThread()
             }
         }
+        searchChat.visibility=View.VISIBLE
+        fabsetting.visibility=View.VISIBLE
+
+
     }
 
     override fun onTouchEvent(event: MotionEvent): Boolean {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -326,6 +328,8 @@ class ChatActivity : AppCompatActivity(), IChatView {
             chatPresenter.stopHotwordDetection()
         }
         hideSoftKeyboard(this, window.decorView)
+        searchChat.visibility = View.INVISIBLE
+        fabsetting.visibility = View.INVISIBLE
```
liveHarshit: Why not `View.GONE`, so it will not accessible.

app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -334,6 +338,8 @@ class ChatActivity : AppCompatActivity(), IChatView {
 
     //Replies user with Speech
     override fun voiceReply(reply: String, language: String) {
+        searchChat.visibility = View.VISIBLE
+        fabsetting.visibility = View.VISIBLE
```
liveHarshit: And what about it?

atm1504: > And what about it?

When we open up sttfragment, and give some voice commands, the chat activity is not again restarted. It simply calls the function voiceReply. So, it was necessary to make the visibility here

Issue #2241, opened 2019-06-01
Voice commands list in STTFragment is not properly responding to clicks


atm1504:
**Describe the bug**
The sample voice commands items in the STT fragment is not responding to clicks simply. Currently, the commands are not every time clickable. It's not accepting the user response. It should respond to the clicks immediately and lead to chat activity.

**Expected Behaviour**
The items should be properly clickable. As soon as it is clicked, the page should redirect to the chat activity also.

**Steps to reproduce it**
Click the mic icon. Press any voice command. You will find your command is not accepted by the app

**Smartphone Info:**

|               | |
|---------------|-|
|Device         | Redmi Note 5 Pro |
|Android Version| Oreo 8.1    |

**Would you like to work on the issue?**
Yes

atm1504:
Should be solved along with https://github.com/fossasia/susi_android/issues/2241

PR #2248, opened 2019-06-04
fix: Fix voice commands actions in voice prompt window

### LINKED ISSUE #2171: Go to chat Activity as soon as voice command is pressed
**Describe the bug**
Currently, the voice commands even after clicked waits until the speech recognition timesout.
This shouldn't happen. It should go to the ChatActivity otherwise it doesn't justify the purpose of these commands. 

**Expected Behaviour**
The voice commands should go to chatAcitivity and display results as soon as they are clicked.

Yes

atm1504:
Fixes #2171 #2241 

Screenshots for the change: 
ezgif com-video-to-gif

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/VoiceCommandsAdapter.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/VoiceCommandsAdapter.kt
@@ -1,6 +1,7 @@
 package org.fossasia.susi.ai.chat.adapters.recycleradapters
 
 import android.content.Context
+import android.content.Intent
 import android.support.v7.widget.RecyclerView
 import android.view.LayoutInflater
 import android.view.View
@@ -8,6 +9,7 @@ import android.view.ViewGroup
 import android.widget.TextView
 import kotlinx.android.synthetic.main.item_voice_commands.view.voiceCommand
 import org.fossasia.susi.ai.R
+import org.fossasia.susi.ai.chat.ChatActivity
 import org.fossasia.susi.ai.chat.ChatPresenter
 import org.fossasia.susi.ai.chat.contract.IChatPresenter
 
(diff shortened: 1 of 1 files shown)


atm1504:
@arundhati24 @batbrain7 @iamareebjamal please review it.

iamareebjamal:
First of all, tell me who has told you that restarting an activity with FLAG_ACTIVITY_CLEAR_TOP is a good idea and solution of the bug. From next time on, if your solution involves FLAG_ACTIVITY_CLEAR_TOP, just don't send the PR and think of other solution

iamareebjamal:
This is like "Oh, I have to show a new message SUSI sent from the server and it is not displaying. Let me just restart the activity so that all messages are refetched from the server and the new message will be displayed. Easy Peasy"

What a terrible solution

Revert all PRs you have made involving FLAG_ACTIVITY_CLEAR_TOP

atm1504:
@iamareebjamal @arundhati24 @batbrain7 please review it. Commit 51f8b67 is a different method that has been used and it's working fine.

app/src/main/java/org/fossasia/susi/ai/chat/adapters/recycleradapters/VoiceCommandsAdapter.kt
```diff
@@ -32,13 +37,16 @@ class VoiceCommandsAdapter(val items: ArrayList<String>, val context: Context?)
...
+            voiceCommand.setOnClickListener {
+                val message = items[adapterPosition]
+                val this_activity = context
+                (this_activity as ChatActivity).setText(message)
+                this_activity.searchChat.show()
+                this_activity.voiceSearchChat.show()
+                this_activity.fabsetting.show()
+                this_activity.btnSpeak.isEnabled = true
+                chatPresenter.startHotwordDetection()
+                this_activity.supportFragmentManager.popBackStackImmediate()
```
iamareebjamal: Why not sending to presenter

atm1504: I tried that also. But, what I found that, as we are not starting the chatActivity (none of the following actions onResume, onPause, onCreate gets executed), so the presenter is failing to control any UI changes of the chatActivty. Hence, that method failed. Then, I found out this method useful.

arundhati24: This doesn't seem to be a good way to set the click listener. To keep things clean, you should create and pass an item click listener (that requires adapter position and the message) from the view to the adapter, while initialising the adapter, and inside the init of view holder provide the adapter position and message to be set. Then, from your fragment, tell your presenter to start hotword detection and once that is done successfully, let the presenter tell the view to display any changes, that are required, to the user.

atm1504: @arundhati24 I even tried that method. Whats, the issue here is that when we do popBackStackImmediate(), the ChatActivity is opened. The issue here is, the state of ChatActivity is neither onPause, onResume or onCreate. So, even if I try to do anything using view, it simply fails. Those options that have onClick function in XML only work. Even, the chat presenter doesn't get access to the view in such cases(I have checked it). That was the main reason, I came up with this kind of solution, after so many tries. I also found, that the one who created the files of STT Fragment, also had a similar mindset.

PR #2265, opened 2019-06-14
fix: Convert server datetime response to local

### LINKED ISSUE #2106: Time is not properly displayed
**Describe the bug **
When asked susi for time right now,  it shows correct  and date but time is not appropriate. 
** Expected behaviour **
The time is currently 18.39 according to 24 hour clock but it shows 2.03 pm
**Screenshots of the issue**
Screenshot_20190329-183721_SUSIAI
Yes

atm1504:
Fixes #2106 

Changes: 
Convert the CET time zone response to the local timezone of the device

Screenshots for the change: 
Screenshot 2019-06-15 at 7 46 11 PM

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
@@ -20,6 +20,11 @@ import org.fossasia.susi.ai.data.model.ChatMessage
 import org.fossasia.susi.ai.helper.Constant
 import org.fossasia.susi.ai.rest.ClientBuilder
 import org.fossasia.susi.ai.rest.responses.susi.SkillRatingResponse
+import org.threeten.bp.Instant
+import org.threeten.bp.ZoneId
+import org.threeten.bp.ZonedDateTime
+import org.threeten.bp.format.DateTimeFormatter
+import org.threeten.bp.format.DateTimeFormatterBuilder
 import retrofit2.Call
 import retrofit2.Callback
 import retrofit2.Response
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/susi/ai/chat/adapters/viewholders/ChatViewHolder.kt
```diff
@@ -123,6 +131,70 @@ class ChatViewHolder(view: View, clickListener: MessageViewHolder.ClickListener,
         }
     }
 
+    @SuppressLint("NewApi")
+    private fun getDateTime(answer: String): Any {
```
iamareebjamal: Use a library https://github.com/JakeWharton/ThreeTenABP

atm1504:
Apk: 
[app-fdroid-debug.apk.zip](https://github.com/fossasia/susi_android/files/3293178/app-fdroid-debug.apk.zip)

atm1504:
> You're still writing your own logic

You mean simply by using this library we can directly convert the CEST format to our local system. But probably there will be an issue. As the response that we receive contains "CEST", but the actual timezone name is CET. So, we cannot directly convert the response. We have to convert it into a suitable form that our parser can understand, then only we can convert. This was the issue that I faced for a few days. Upon searching I found this kind of solution as a reliable one.

iamareebjamal:
What's the use of using the library when you're writing your entire logic?

iamareebjamal:
Write the exact response format received from the server

atm1504:
WhatsApp Image 2019-06-15 at 11 04 35 PM
This is the format in which we receive our data from the server.

atm1504:
@iamareebjamal shall I directly put the response that I receive from the server into the following library function and tell it to give output in the local timezone?

atm1504:
> Write the exact response format received from the server

I tried to parse directly the exact format received from the server using the library. But the library failed to parse it. This shows that the library doesn't support the format we receive the data. Please put your views.

iamareebjamal:
Write the exact response format received from the server

atm1504:
> Write the exact response format received from the server
> 
> If the server doesn't send timestamp or ISO format string, open the issue on server

I tried to replicate the way you said. But it failed. The response is not in ISO format, not even its a timestamp. If we try to solve from the app, we need to write down the logic. Else we have to create an issue in the server. What should I do know?

iamareebjamal:
Create issue on server

PR #2279, opened 2019-06-25
feat: Add location of the device

### LINKED ISSUE #2277: Add the gps location of the smart speaker to the server
Currently, when we configure our susi smart speaker and add that device to the server, we simply use dummy geo location. What we want is that we should actually add the current position of the device as the geolocation. Get this position via GPS.

**Additional context**
The app already has access to GPS, what we need is to get its position values and add it to the server instead of using dummy values.

Yes

atm1504:
Fixes #2277 

Changes: 
The actual location of the device is added to the server.

Screenshots for the change: 
Screenshot 2019-06-25 at 8 11 21 PM
ezgif com-video-to-gif


atm1504:

atm1504:
Apk: 
[app-fdroid-debug.apk.zip](https://github.com/fossasia/susi_android/files/3325889/app-fdroid-debug.apk.zip)

PR #2283, opened 2019-06-27
feat: Delete feedbacks of skills

### LINKED ISSUE #1473: Add option to allow the user to edit and delete feedback
**Expected Behaviour**

Add an option to the view holder of the user's feedback to allow him to edit or delete his feedback.

Yes

atm1504:
Fixes #1473

Changes:
Users can now delete feedbacks

Screenshots for the change: 
To be updated after completion.


atm1504:
Screenshot: 
Screenshot 2019-06-27 at 10 27 35 PM

ezgif com-video-to-gif

Apk: 
[app-fdroid-debug.apk.zip](https://github.com/fossasia/susi_android/files/3335879/app-fdroid-debug.apk.zip)

atm1504:
@iamareebjamal @batbrain7 @arundhati24 please review it.

iamareebjamal:
Nothing should be relaunched on deleting the feedback. What a terrible UX

atm1504:
> Nothing should be relaunched on deleting the feedback. What a terrible UX

Removing it

app/src/main/java/org/fossasia/susi/ai/rest/responses/susi/GetSkillFeedbackResponse.kt
```diff
@@ -10,7 +10,7 @@ data class GetSkillFeedbackResponse(
     val accepted: Boolean = false,
     val message: String = "",
     @SerializedName("feedback")
-    var feedbackList: List<Feedback> = ArrayList(),
+    var feedbackList: ArrayList<Feedback> = ArrayList(),
```
iamareebjamal: Why?

atm1504: What I want is that when we delete the feedback, what we should do is that we should remove that item from the list so that the feedback is not shown to the user. It has been removed from the server. To reload and show the fresh response from the server, we need to relaunch the page. So, to avoid that it would be better to just remove that item from the list. And display the rest items.

iamareebjamal: Your response makes no sense. Why change List to ArrayList

atm1504: Because of the presence of List type, it was not allowing operations like remove, removeAt in the feedbackList. SUch operations are possible in ArrayList type only. So, I changed it.

iamareebjamal: Use MutableList

app/src/main/java/org/fossasia/susi/ai/skills/skilldetails/SkillDetailsFragment.kt
```diff
@@ -578,4 +578,12 @@ class SkillDetailsFragment : Fragment(), ISkillDetailsView {
         skillDetailsPresenter.onDetach()
         super.onDestroyView()
     }
+
+    override fun onResume() {
+        if (PrefManager.getBoolean(R.string.is_feedback_deleted, false)) {
```
iamareebjamal: Why?

atm1504: If a user deletes his or her feedback in the allReviewsActivity, then also he is able to do so. It is deleted in the server. Also, we remove it from the local list that contains the list of the data. Now, if user press backs, his feedback would be shown. This is because we are not fetching the data again. And in the allReviewsActivity we deleted in the variable of that file along with the server. It has not been deleted in the skillDetailsFragment. So, I used this preference to know if the user has deleted his feedback in allReviewActivity or not. If he has done so, then we are reloading the feedbacks only once again. Else, we are not making any API calls.

iamareebjamal: Terrible solution. @liveHarshit Please tell what is done in normal apps where we delete something and need to show updated list

atm1504: ezgif com-video-to-gif

Let me explain to u in detail why I did so.
Clearly, in the video, u can see there is an option "See All Reviews", pressing it redirects to a new page "AllReviewsActivity". In the intent, the feedbackList is passed as an extra to the intent. Now, there might be a situation as I showed u the gif that users delete his feedback while he/she is in "AllReviewsActivity". When he does so, we are also deleting it from the local list of the page also in the server. Now, the user will surely press back-button. If, he does so, the
"SkillDetailsFragment" is popped back. We don't refresh any data if the user does show. Now, as the user has already deleted his feedback, it should not appear back in the "SKillDetailsFragment". Now, the question arises who to know whether the user has deleted the feedback while he was in "AllReviewsActivity". To know that I used the preference manager. If the user has deleted the item we have 2 options to show the changes to the user. One is to delete the item from the feedback list and show it to the user. The other one is to reload the feedback list back from the server. In the second option, only the feedback list is reloaded not the other data of the page. A similar approach was already in use in the app, where if the feedback is added/updated by the user the feedback list is reloaded.

@iamareebjamal @liveHarshit please provide your views now.

iamareebjamal: > Now, the question arises *how* to know whether the user has deleted the feedback while he was in "AllReviewsActivity"

Very simple, it's deleted from the database so it won't be there

atm1504: We don't have any database setup for storing the feedbacks. It's simply stored into variables.

iamareebjamal: Then use startActivityForResult, not shared preferences

atm1504: okk

atm1504: > Then use startActivityForResult, not shared preferences

I don't think it can be done. Because the above situation(using preference manager) was done to handle if changes occur in another file and after that user presses back. Now, if we try to go for onActivityForResult when the user deletes the feedback on allReviewActivity page, aren't we forcing the user to go back(I mean we are performing the back action in code)?

atm1504: @iamareebjamal I tried implementing it. But the scene here is different from the normal cases. I don't think it is feasible to use startActivityForResult here.

iamareebjamal: How is it different. You want to know if the child activity has deleted certain reviews. It's perfect

atm1504:

iamareebjamal:
@liveHarshit Please review

app/src/main/java/org/fossasia/susi/ai/rest/services/SusiService.kt
```diff
@@ -205,6 +206,18 @@ interface SusiService {
...
+    /**
+     * Delete Feedback posted by the user
+     * @param query A query map consisting of following key value pairs
+     * model       Model of the skill (e.g. general)
+     * group       Group of skill (e.g. Knowledge)
+     * language    Language directory in which the skill resides (e.g. en)
+     * skill       Skill Tag of object in which the skill data resides
+     * access_token Access token of the user
+     */
+    @GET("/cms/removeFeedback.json")
```
liveHarshit: Why GET? As you have written `Delete Feedback posted by the user` then it should be POST

atm1504: The api accept get method. Moreover it was mentioned in gitter channel to make get request. Also get method is also present during addition of feedback. So I used it.

iamareebjamal: @liveHarshit Yeah, SUSI server is weird that way. Instead of /api/v1/whatever_stuff, it is /aaa/whateverStuff.json

No idea why it's `aaa` or .json after every URL. 

Or why is data even sent as GET and not POST. 

Or why is access token passed as a GET parameter and not in header. Freakishly insecure way of doing things

liveHarshit (approved):
Honestly, I don't understand what actually going on. Why you need to create a new class for deleting feedback. It should directly implemented from the server side. Looks good to me, code wise. :+1:

atm1504:
> Honestly, I don't understand what actually going on. Why you need to create a new class for deleting feedback. It should directly be implemented from the server side. Looks good to me, code wise. 👍

I agree with you. I also faced a lot of difficulties adapting myself to this bit weird kind of coding style present in this app and thus committed too many mistakes before. Never worked on such kind of app before.

mariobehling:
This issue was not in the milestone. We currently dont have the focus to work on this. We need to align all clients and no-one is working on this issue you took up here. We discussed at the beginning of GSoC that the smart speaker needs to work first before anything else. You still dont put your entire energy on that. Closing this as it is currently not in the scope of the milestones.

atm1504:
@mariobehling I think we should reopen this now. As most of the planned feature for smart speaker has been added. The few left needs the support of susi_installer. But, those features like (music_control) has been commented out in the installer. So, I am unable to work on them now.

PR #2310, opened 2019-07-18
chore: Update shimmer dependency

atm1504:
Fixes #2309 

Changes:
* Updated the shimmer version
* shifted the value of version to the config file

Screenshots for the change: 
No relevant changes


app/src/main/java/org/fossasia/susi/ai/chat/ChatActivity.kt
```diff
@@ -620,4 +620,4 @@ class ChatActivity : AppCompatActivity(), IChatView {
         Timber.d(videoId)
         youtubeVid.playYoutubeVid(videoId)
     }
-}
+}
```
iamareebjamal: Revert

PR #2317, opened 2019-07-22
fix: Remove filter options from settings page

### LINKED ISSUE #2315: Filter option appears in settings page
**Describe the bug**
Filter options are visible even in settings fragment. It should not be visible there. Just need to correct it onPrepareMenuOptions. Probably the bug was created because visibility of it in settings page was not removed. It is working fine in other places.

**Steps to reproduce it**
* Open the skills page (No filter option is visible)
* Now, open settings.

**Screenshots of the issue**
Screenshot 2019-07-22 at 6 50 34 PM

Yes

atm1504:
Fixes #2315

Changes: 
* Fixed the visibility

Screenshots for the change: 
Screenshot 2019-07-22 at 7 02 31 PM


atm1504:

PR #2476, opened 2020-01-31
fix: Added unit test for datahelper class

### LINKED ISSUE #2100: Unit tests for feedback activity
**Describe the bug**
There is no unit tests written for feedback activity

**Expected Behaviour**
There has to be unit tests on feedback activity as well using Espresso for faster results

Yes

pallav12:
Fixes #2100 

Changes: Made feedback activity testable and added unit test, since only testable unit in Feedback activity is format date. So migrated getDate() function to DateTimeHelper and added unit test for all other functions of DateTimeHelper

Screenshots for the change: 
Screenshot (8)

### DIFF
--- a/app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
+++ b/app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
@@ -39,6 +39,11 @@ object DateTimeHelper {
             return sdf.format(Date())
         }
 
+    val today: Date
+        get() {
+            return Calendar.getInstance().time
+        }
+
     /**
      * Method to format date from server
 
@@ -46,7 +51,7 @@ object DateTimeHelper {
      * *
      * @return Date
      */
-    private fun formatDate(date: String): Date? {
+    private fun formatDate(date: String): Date {
         val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
         sdf.timeZone = TimeZone.getTimeZone("UTC")
         var dateIn: Date?
@@ -54,7 +59,7 @@ object DateTimeHelper {
             sdf.parse(date)
         } catch (e: Exception) {
             Timber.e(e)
-            null
+            today
         }
 
         return dateIn
@@ -85,6 +90,25 @@ object DateTimeHelper {
         val sdf = SimpleDateFormat("hh:mm aaa")
         val tz = TimeZone.getDefault()
         sdf.timeZone = tz
-        return sdf.format(formatDate(date))
+        return if (date.isEmpty()) {
+            throw IllegalArgumentException("date argument is empty")
+        } else {
+            sdf.format(formatDate(date))
+        }
+    }
+
+    fun formatDate(timestamp: String, months: Array<String>): String {
+        return if (timestamp.length > 10 && months.size == 12) {
+            var date: String? = ""
+            timestamp.trim()
+            val month = timestamp.substring(5, 7).toInt()
+            date = timestamp.substring(8, 10) + " " +
+                    months[month - 1].toString() +
+                    ", " + timestamp.substring(0, 4)
+
+            date
+        } else {
+            throw IllegalArgumentException("Timestamp or Months format not correct")
+        }
     }
 }
--- a/app/src/test/java/org/fossasia/susi/ai/helper/DateTimeHelperTest.kt
+++ b/app/src/test/java/org/fossasia/susi/ai/helper/DateTimeHelperTest.kt
@@ -1,0 +1,71 @@
+package org.fossasia.susi.ai.helper
+
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.TimeZone
+import junit.framework.TestCase.assertEquals
+import org.junit.Test
+
+class DateTimeHelperTest {
+    private val TIMEZONE_SINGAPORE = "Asia/Singapore"
+    private val TIMEZONE_US_PACIFIC = "US/Pacific"
+    private val TIMEZONE_SYDNEY = "Australia/Sydney"
+    private val TIMEZONE_AMSTERDAM = "Amsterdam"
+    private val TEST_TIME = "2020-01-30T12:18:37.262Z"
+
+    val months = arrayOf("Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Aug", "Nov", "Dec")
+    val emptyMonths = arrayOf("")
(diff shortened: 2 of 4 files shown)


app/src/test/java/org/fossasia/susi/ai/helper/DateTimeHelperTest.kt
```diff
@@ -0,0 +1,17 @@
...
+class DateTimeHelperTest {
+    @Test
+    fun getDateTest() {
+        Assert.assertEquals(" Jan 30, 2020", DateTimeHelper.getDate("2020-01-30T12:18:37.262Z"))
+    }
+
+    @Test
+    fun formatDateTest() {
+        val countries = arrayOf("Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Aug", "Nov", "Dec")
+        Assert.assertEquals("30 Jan, 2020", DateTimeHelper.formatDate("2020-01-30T12:18:37.262Z", countries))
```
iamareebjamal: Add  case for null timestamp as well

pallav12: @iamareebjamal added null checks as well as tests for null and empty values. One question though, I tried adding test for one remaining function ie getTime() but every time CI build is failing. That's because i have to hardcode expectation for input, but expectations would be different depending upon different timezone(github vs my), so according to me it's not possible... or is it??

iamareebjamal: See tests for  open-event-organizer app

pallav12: Thanks... that worked, implemented tests for different timezones

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
```diff
@@ -71,7 +71,11 @@ object DateTimeHelper {
         val sdf = SimpleDateFormat(" MMM dd, yyyy")
         val tz = TimeZone.getDefault()
         sdf.timeZone = tz
-        return sdf.format(formatDate(date))
+        if (formatDate(date) == null || sdf.format(formatDate(date)) == null) {
+            return ""
+        } else {
+            return sdf.format(formatDate(date))
+        }
```
iamareebjamal: If this is needed, then the argument should be made non nullable

pallav12: Made date non nullable so this argument isn't needed any more, returning today date when cannot parse ie when null. please recheck...

iamareebjamal: Then remove this if else

pallav12: removed this, and updated DateTimeHelperTest accordingly

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
```diff
@@ -85,6 +90,25 @@ object DateTimeHelper {
         val sdf = SimpleDateFormat("hh:mm aaa")
         val tz = TimeZone.getDefault()
         sdf.timeZone = tz
-        return sdf.format(formatDate(date))
+        return if (date.isEmpty()) {
+            ""
```
iamareebjamal: Remove this

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
```diff
@@ -85,6 +90,25 @@ object DateTimeHelper {
...
+            var date: String? = ""
+            timestamp.trim()
+            val month = timestamp.substring(5, 7).toInt()
+            date = timestamp.substring(8, 10) + " " +
+                    months[month - 1].toString() +
+                    ", " + timestamp.substring(0, 4)
+
+            date
+        } else {
+            ""
```
iamareebjamal: Remove this

pallav12: What exactly?? timestamp!=null can be removed by making timestamp non nullable, if else condition can't be removed because if not checked it'll expose indexOutOfBound exceptions in months and string length(So if this happens just return empty string ie don't print anything in date).

iamareebjamal: Throw error instead

pallav12: Thrown error and added corresponding tests, There wont be these redundancies from next time.
please notice updated screenshot also.

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
iamareebjamal: This is left

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
```diff
@@ -85,6 +90,25 @@ object DateTimeHelper {
...
+            var date: String? = ""
+            timestamp.trim()
+            val month = timestamp.substring(5, 7).toInt()
+            date = timestamp.substring(8, 10) + " " +
+                    months[month - 1].toString() +
+                    ", " + timestamp.substring(0, 4)
+
+            date
+        } else {
+            throw IllegalArgumentException()
```
iamareebjamal: Pass proper message in the exception

app/src/main/java/org/fossasia/susi/ai/helper/DateTimeHelper.kt
```diff
@@ -85,6 +90,25 @@ object DateTimeHelper {
...
         sdf.timeZone = tz
-        return sdf.format(formatDate(date))
+        return if (date.isEmpty()) {
+            throw IllegalArgumentException()
+        } else {
+            sdf.format(formatDate(date))
+        }
+    }
+
+    fun formatDate(timestamp: String?, months: Array<String>): String? {
```
iamareebjamal:
```suggestion
    fun formatDate(timestamp: String, months: Array<String>): String {
```

pallav12:
Thanks @iamareebjamal  for having patience... I'll try to avoid any redundancies from now.

iamareebjamal (approved):

PR #2533, opened 2020-09-13
feat: Fast Image Loading 

FareesHussain:
Changed Picasso Image Loader with Glide image loader 
Reasons - 
1. Glide’s loading times are faster and it uses a small amount of memory for cache
2. Glide might be a better alternative to Picasso when memory footprint is less of a concern or more and larger images need to be processed


FareesHussain:
i can't understand why i'm getting the link check error in the circle-ci 
can some one please help with this

